PR #62, opened 2026-01-22
Implement auto-connect feature for last session

JanRi3D:


andreknieriem:
That is very very great! Can you look at the findings from CoPilot? I would like to merge this for 1.10.0 today!

JanRi3D:
> That is very very great! Can you look at the findings from CoPilot? I would like to merge this for 1.10.0 today!

Can do!

PR #67, opened 2026-01-23
feat: Brazilian Portuguese translation

### LINKED ISSUE #65: Brazilian Portuguese translation
If you'd like, and if it's not too difficult, I think I could work on a PT-BR translation for the app, if there isn't one already.

MicaelJarniac:
Closes #65


MicaelJarniac:
I've asked GPT 5.2 to translate it, then went over it and tweaked some things. I think it's alright. I'm a bit unsure about the "soft" buttons on the key remapping section, though, but I think it's fine as is.

andreknieriem:
Thank you very much!

PR #70, opened 2026-01-24
Vibe-crash-fix

MicaelJarniac:
Summary

Prevented stale connection attempts from starting transport by tracking a monotonically increasing connection attempt ID and invalidating it on disconnect.
Hardened handshake flow: added retries with backoff for the initial version request/response and clearer logging for failures.
Added a diagnostic warning when the connection is cleared before transport start to confirm the race condition.
Details

AapService.kt
Track connection attempts with AtomicInteger and ignore stale results.
Invalidate current attempt on disconnect.
Snapshot and log when accessoryConnection is cleared before transport start.
AapTransport.kt
Retry the version request/response up to 3 times with short delays.
Improve logs for send/receive failures and final retry exhaustion.
Why

Java SSL handshake failures were often caused by transient USB connection instability (sendBlocking returning -1) and overlapping connection attempts. The new logic avoids races and tolerates short re‑enumeration windows.


andreknieriem:
So, did you test this? If not, should I build an apk and you test it?

MicaelJarniac:
> So, did you test this? If not, should I build an apk and you test it?

I have not tested it, nor do I know how to build the APK.

If you could build it for me, I'd gladly test it.

PR #73, opened 2026-01-24
feat: Add Polish translation for app strings

Kacper1263:
Hi, I've created and tested Polish translation

obraz


andreknieriem:
Thank you!

PR #78, opened 2026-01-25
Create strings.xml with russian translation

prostozema:
translate strings.xml from english to russian using deepseek and made some adjustments


andreknieriem:
Thank you!

PR #95, opened 2026-02-01
Fullscreen behaviour not refreshed after applying the setting and KitKat compatibility issues

nicoruy:
Because the settings screen is a fragment, not an activity, the onCreate or onResume methods are not called when returning to the main screen. As a result, setFullscreen is not called, and the screen behavior is not updated until the app is restarted. 
Changing settings to an Activity fixes this problem. 
Also, the SurfaceActivity class was changed to inherit from AppCompatActivity instead of Activity, and some conditional logic was added to SystemUI to address compatibility issues with KitKat.


andreknieriem:
Hey Nico, thanks for your PR. It has a lot of changes and I need some time to test it all. The topic with the fullscreen and other flags was very a pain. I need to test it on several devices first. Thanks a lot for your work 👍

nicoruy:
My pleasure! 
I discovered another related bug when the screen has a non-standard aspect ratio. It results in black bars on the sides, and touch input becomes misaligned with the rendered video. I already have a fix and will submit a PR after a few more tests.

andreknieriem:
> My pleasure!
I discovered another related bug when the screen has a non-standard aspect ratio. It results in black bars on the sides, and touch input becomes misaligned with the rendered video. I already have a fix and will submit a PR after a few more tests.

No way! What device do you use? The way the whole scaling and so works is, that AA only have several resolution. Eg 1920x1080px. If you have a Device with 1920x720px my code chooses 1080p and sends margin 360px margin to the phone. This is the way AA known in which space to render its content. After that the view in the headunit scales to the right points to only see the filled area and not the empty black areas of the stream. It works for me on all my emulators and devices, even in Portrait mode. So it is very interesting, that you experienced this behavior.

PR #120, opened 2026-02-11
Add dynamic language selector to app settings

andrecuellar:
## Features:
- Add language selector in General settings that allows users to choose their preferred app language independently from the system language
- Dynamically detect available languages from res/values-* directories using AssetManager.getLocales() - new translations automatically appear without code changes
- Store user preference in SharedPreferences and apply at runtime
- Include "System Default" option that uses Android system language
- Apply locale to all activities via BaseActivity.attachBaseContext()
- Restart activity when language is changed to apply new locale

This allows users with Chinese head units to keep the system in English [because the translation of some apps is terrible] while using the app in their native language (e.g., Spanish).

PR #121, opened 2026-02-12
Fix/hardcoded strings

andrecuellar:


andrecuellar:
Hi @andreknieriem 

I fixed multiple hardcoded strings by adding proper string resources with English and Spanish translations.

Note: This PR depends on #120 (language selector) being merged first, as it builds on top of those changes.

PR #124, opened 2026-02-13
Native aa

andreknieriem:


saksonovdev:
mega bombastisch

andreknieriem:
Thank you, but there is a lot of work to do and it will only work on a very limited amount of devices.

andreknieriem:
I will bring this back soon, but I need to rewrite this from scratch because of CommManager etc.

PR #127, opened 2026-02-13
Fix audio truncation

nicoruy:
Audio of driving instructions sometimes gets truncated before finishing. This happens because the buffer gets flushed before it is fully processed. 
This fix reads the buffer if there is still data, even when it isRunning in false. Cleanup is also improved to handle the remaining data and close gracefully.

I also made a small change to the build.gradle script to use passwords from a secrets.properties file, which is excluded from git, instead of environment variables. If the file is not present, it falls back to environment variables.


andreknieriem:
Thank you! This is great

PR #129, opened 2026-02-15
Correction for Chinese Mediatek Headunit Connection

tsabaia:
An error(System.out: e:java.lang.ClassNotFoundException: com.mediatek.cta.CtaHttp W System : ClassLoader referenced unknown path: system/framework/mediatek-cta.jar) generated by some headunit devices with a Mediatek processor prevents connection to Android Auto. The added code prevents this error from dropping the connection.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/connection/SocketAccessoryConnection.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/connection/SocketAccessoryConnection.kt
@@ -98,7 +98,18 @@ class SocketAccessoryConnection(private val ip: String, private val port: Int, p
                 }
                 
                 transport.soTimeout = 15000
-                transport.connect(InetSocketAddress(ip, port), 5000)
+                // Chinese Headunit Mediatek Correction
+                try {
+                    transport.connect(InetSocketAddress(ip, port), 5000)
+                } catch (e: Throwable) {
+                    val errorMessage = e.message ?: e.toString()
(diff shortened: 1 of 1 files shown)


andreknieriem:
So the try prevents the error, but does the connection work?

tsabaia:
This code change solved the connection problem.

I opened and closed it several times to test, and the connection worked perfectly.

Those logs I sent you on XDA had several lines with the error `System.out: e:java.lang.ClassNotFoundException: com.mediatek.cta.CtaHttp W System : ClassLoader referenced unknown path: system/framework/mediatek-cta.jar`

From what Gemini explained, every time this error occurred, the connection was broken, and that's why I could never use Android Auto.

This code change makes the app ignore this error and not break the connection as it did before. This way, Android Auto connects normally.

andreknieriem:
That's great! I will merge it as soon as. I'm back at work.

tsabaia:
I have a video saved here from before this code change. You can see that the phone showed a connection signal, but nothing appeared on the head unit.

https://www.youtube.com/shorts/w0RUcmwHz8A

After changing the code, the connection works perfectly.

https://www.youtube.com/shorts/bi_4l6oD_uI

andreknieriem:
Why closed?

tsabaia:
I have no experience with Github, I didn't know that deleting the Branches affects the pull request.

I opened a new Pull Request. https://github.com/andreknieriem/headunit-revived/pull/134

PR #142, opened 2026-02-19
Fix: return to active Android Auto session when app is reopened

andrecuellar:
## Problem
When Android Auto is running (`AapProjectionActivity` visible) and the user minimizes the app then taps the launcher icon, `MainActivity` is brought to front via its `singleTask` launch mode. Neither `onCreate()` nor `onResume()` check if there's an active AAP session, so the user sees the home screen instead of the running Android Auto projection. The connection is still alive (green status in USB list), but there's no way to get back to it without disconnecting and reconnecting.

Additionally, tapping a connected (green) device in `UsbListFragment` creates a bare `Intent` without proper flags, which may not bring the existing `AapProjectionActivity` instance to front correctly.

## Summary
- When an AA session is running and the user minimizes then taps the launcher icon, `MainActivity` now detects the active connection in `onResume()` and brings `AapProjectionActivity` back to front
- Fixes the USB list connected-device tap to use `AapProjectionActivity.intent()` factory method with `FLAG_ACTIVITY_REORDER_TO_FRONT` so the existing projection instance is reused

## Test plan
- [ ] Connect phone via USB, start Android Auto session
- [ ] Press Home to minimize, then tap launcher icon — should return to AA projection
- [ ] Verify normal connection flow still works when no session is active

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/main/MainActivity.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/main/MainActivity.kt
@@ -13,6 +13,7 @@ import androidx.activity.viewModels
 import androidx.core.app.ActivityCompat
 import androidx.core.content.ContextCompat
 import com.andrerinas.headunitrevived.R
+import com.andrerinas.headunitrevived.aap.AapProjectionActivity
 import com.andrerinas.headunitrevived.aap.AapService
 import com.andrerinas.headunitrevived.app.BaseActivity
 import com.andrerinas.headunitrevived.utils.AppLog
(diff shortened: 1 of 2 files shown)


andreknieriem:
This with https://github.com/andreknieriem/headunit-revived/pull/140 should do the trick. The notification had an error too

andreknieriem:
I've added your code to the 2 files. This PR can be closed :) Thank you so much for your input!

andrecuellar:
> I've added your code to the 2 files. This PR can be closed :) Thank you so much for your input!

@andreknieriem Thanks! I'm at the office, so I can't test it yet. That's why I left it in draft, but I hope it worked! I'll test it in a couple of hours.

andrecuellar:
@andreknieriem It works!

Thank you so much for quickly implementing this improvement; the user experience has improved so much this month. What a great project!

andreknieriem:
> That is so true. The app is only 2 month on github and it changes so much and it gets better and better! The first 4 Month of bringing it back to live are not included :D

Issue #161, opened 2026-02-27
Intermittent WiFi Hotspot Connection Issue – Requires Clearing Cache to Reconnect


sequlla:
I am using Headunit Revived with my phone’s **WiFi** **hotspot** for wireless Android Auto connection.

In most cases, it works fine — about 7 out of 10 times it connects successfully without any issues. However, after several successful connections, it suddenly stops connecting completely. The head unit and phone can see each other, but Android Auto does not start.

The only workaround I have found is to **clear** **the** **cache** of both the Headunit Reloaded app and the Android Auto app. After clearing the cache, the connection works perfectly again for several attempts — and then the issue repeats.

This behavior is consistent and reproducible.

**Device**: Motorola Edge 30 Fusion
**Connection** **method**: WiFi hotspot (phone as hotspot)
**Android** **version**: Android 14

Is this a known issue? Is there a permanent fix so that clearing cache is not required?

andreknieriem:
This is a known bug since forever and it gets better and better the last 2 month (HURev is still too new :D) . But the reason why this happens is still not clear. You don't need to clear caches. If you disable and enable wifi on your phone it might work too. The problem in the past was, if you disconnect hard or connect with wrong settings (eg h265 codec, if your device only can h264, wrong resolution etc) the phone/Android Auto denies any further connection after some attempts and for a several time. 

I will let this issue open to find the last bugs :) Thank you for opening this issue.

sequlla:
> This is a known bug since forever and it gets better and better the last 2 month (HURev is still too new :D) . But the reason why this happens is still not clear. You don't need to clear caches. If you disable and enable wifi on your phone it might work too. The problem in the past was, if you disconnect hard or connect with wrong settings (eg h265 codec, if your device only can h264, wrong resolution etc) the phone/Android Auto denies any further connection after some attempts and for a several time. 
> 
> I will let this issue open to find the last bugs :) Thank you for opening this issue. 

I've tried everything.  The only solution is that I have to stop the android auto server on the phone, clear the cache, turn off headunit revived on the tablet and clear the cache on it and only then can I use it normally again.  If I skip something in this process it will never connect.

andreknieriem:
The Android Auto Server on the phone is a whole other thing. That thing drives me crazy in testing local between 2 emulators. It is much better to use wireless helper and never touch the android auto server again :D 

https://play.google.com/store/apps/details?id=com.andrerinas.wirelesshelper
[...]

PR #160, opened 2026-02-27
Add retry button to stuck loading screen

andrecuellar:
## Summary

The "Android Auto is starting..." loading screen in `AapProjectionActivity` sometimes gets stuck indefinitely (~10% of the time). This happens when the connection handshake succeeds but no video frames arrive — due to a slow USB dongle, phone-side stall, or dropped video config messages. The existing `videoWatchdogRunnable` requests keyframes every 3s but never gives up or offers recovery UI.

This PR adds a **Retry button** that appears after 10 seconds on the loading screen, giving the user a way to recover without physically disconnecting.

### Retry button on stuck loading screen
- Restructured `loading_overlay` layout to include a retry button (initially hidden)
- 10-second timer (`retryTimerRunnable`) shows the retry button if loading overlay is still visible
- When first video frame arrives, retry button is hidden and timer is cancelled
- `showRetryDialog()` with two options:
  - **"Retry Android Auto"** — soft retry, disconnects and reconnects (always available)
  - **"Reset USB Connection"** — full USB reset with stability-aware polling (USB only)

### Configurable max auto-reconnect attempts
- New `maxAutoRetryAttempts` setting (default 3, range 1-10) with slider in Settings
- Rewritten `retryConnection()` with configurable retry loop:
  - Progressive delays: 1s for first attempt, 3s for subsequent
  - 5-second wait for connection result after each attempt
  - Logs attempt number/max for each try

### Service lifecycle fix during retry
- Set `isConnecting = true` before `retryConnection()` and `resetUsbAndReconnect()` coroutines
- Prevents service destruction during retry delay (service was being destroyed because `onDisconnect()` sets `isConnected = false` and `serviceJob.cancel()` kills the coroutine)
- Added `CancellationException` handling to properly reset `isConnecting` when service scope is cancelled

### Projection activity foreground fix (from #139)
- Changed `AapProjectionActivity` launch flags from `FLAG_ACTIVITY_REORDER_TO_FRONT` to `FLAG_ACTIVITY_CLEAR_TOP | FLAG_ACTIVITY_SINGLE_TOP`
- Applied in both `onConnectionResult()` and the notification content intent
- Ensures the projection screen reliably comes to the foreground on all devices

### Experimental feature warning (from #139)
- Added confirmation dialog when enabling "USB adapter stability check" toggle
- Warns users that the feature is experimental and may not work on all devices
- Includes a hint to first test USB auto-connect with the stability option turned off
- Dialog reverts the toggle if the user cancels

### String resources
[...]

Issue #147, opened 2026-02-21
USB Autostart, doesn't start USB Android Auto Dongle :/ 


mrkontrast-coder:
The USB Autostart function doesn't actually autostart an USB Android Auto Dongle connection, and I still have to go into USB section, and press the detected divace twice.

mrkontrast-coder:
The divace I use actually, connects, and disconnects in a loop. If i go into HUR and make USB connection then it stops and shows Android Auto

mrkontrast-coder:
If I can help somehow with is let me know, I will make logs next time, in my car and pin them here

andrecuellar:
> The USB Autostart function doesn't actually autostart an USB Android Auto Dongle connection, and I still have to go into USB section, and press the detected divace twice.

@mrkontrast-coder 
That's a problem with the dongle; I'm having the same issue with mine.

The problem is that the dongle changes its ID every time it reconnects, so the USB auto-connect feature no longer works, as if it were a different device.

I'm developing a feature that will automatically connect any connected device, so that these types of devices can connect automatically.

brownah:
Would be great if app detects if a device is plugged in and starts android auto automatically without need to tap anything 🙏

mrkontrast-coder:
> > The USB Autostart function doesn't actually autostart an USB Android Auto Dongle connection, and I still have to go into USB section, and press the detected divace twice.
> 
> @mrkontrast-coder 
> That's a problem with the dongle; I'm having the same issue with mine.
> 
> The problem is that the dongle changes its ID every time it reconnects, so the USB auto-connect feature no longer works, as if it were a different device.
> 
> I'm developing a feature that will automatically connect any connected device, so that these types of devices can connect automatically.

That would be amazing. I also noticed that it is different ID every time it connects. Its very frustrating, and I only menaged to automate the startup with Tasker, but its far from convenient, and messes up most times.

Is it gonna be implemented in the next update ?

andrecuellar:
> > > The USB Autostart function doesn't actually autostart an USB Android Auto Dongle connection, and I still have to go into USB section, and press the detected divace twice.
> > 
> > 
> > [@mrkontrast-coder](https://github.com/mrkontrast-coder)
> > That's a problem with the dongle; I'm having the same issue with mine.
> > The problem is that the dongle changes its ID every time it reconnects, so the USB auto-connect feature no longer works, as if it were a different device.
[...]

PR #139, opened 2026-02-19
Add auto-connect when single USB device is available

andrecuellar:
Add a new setting toggle "Auto-connect USB device" that automatically initiates an Android Auto connection when only one compatible USB device is detected, eliminating the need to manually navigate to the USB list and tap the device.

Implementation:
- Add autoConnectSingleUsbDevice boolean setting (default: off)
- Add toggle in Settings under General section
- Add auto-connect logic in HomeFragment as priority 3 (after last-session reconnect and self-mode auto-start)
- Handle both standard devices (switch to accessory mode) and devices already in accessory mode
- Skip silently if no USB permission is granted
- Add translations for all 10 supported locales

### Auto-connect single USB device
- New `autoConnectSingleUsbDevice` setting toggle in General section
- HomeFragment auto-connect logic (priority 3) — connects when exactly one USB device is plugged in
- Handles standard devices (switches to accessory mode) and devices already in accessory mode
- Skips silently if no USB permission is granted

### USB adapter stability check
- New `usbStabilityCheck` toggle and `usbStabilityTimeout` slider (3-50s) in Settings
- Waits for USB device to stay continuously connected before auto-connecting — for wireless AA dongles that re-enumerate multiple times
- Used by both auto-connect and USB reset flows

### Auto-connect settings screen
- Dedicated screen to configure auto-connect methods and their priority order (drag to reorder)
- Shows enabled methods in order: Last Session → Self Mode → Single USB

### Bluetooth auto-start improvements
- Automatically enables Bluetooth when selecting a BT auto-start device
- Improved adapter name display strings

### Race condition fixes (isConnecting guard)
- Moved `isConnecting` AtomicBoolean from instance field to companion object (persists across service recreation)
- Set `isConnecting = true` before all connection attempts (`handleConnectionIntent`, `checkAlreadyConnectedUsb`, `performSingleUsbConnect`)
- Reset `isConnecting = false` in `onDisconnect()`, `onDestroy()`, and `onConnectionResult()` finally block
- Split combined guard check in `checkAlreadyConnectedUsb()` into separate checks with individual log messages
- Added logging at `handleConnectionIntent`, `onUsbAttach`, and `ACTION_CHECK_USB` for race condition diagnosis
- Prevents multiple simultaneous connection attempts from UsbAttachedActivity, AapService.onCreate, HomeFragment, and onUsbAttach

### AOA re-enumeration fix
- Fixed accessory device not being connected after AOA mode switch
[...]

PR #167, opened 2026-03-02
Communication refactor (introduction of CommManager class)

nicoruy:
Hi,
I'm trying to fix some connectivity issues. In particular, a long wait when connecting via USB, with the loading spinner staying for a while (sometimes forever).
In the process, I started refactoring the communication classes to make them easier to follow and prevent some concurrency issues. I created a CommManager class that encapsulates connections, transport handshakes, and send commands.
This class now has the sole responsibility for handling communication (and replaced transport in AppComponent). Transport is now private to this class and not exposed. 
CommManager also exposes connectivity status and handles a state machine for the connection process using Kotlin Flow, which is thread-safe by design. This allowed me to remove intents used to coordinate the connection process and variables such as connectionAttemptId.
Attributes and methods that were connection-specific were moved to CommManager. For example, AapService.isConnected is now part of CommManager and commputed from the current state of the connection

Another change implemented is to split the transport start into two phases. One performs the handshake (and runs in parallel with the ProjectionActivity creation), and a second phase starts reading from the transport only when the surface is available and ready, preventing the first frame from being missed.

During the process, I fixed some bugs and added a ClientSessionContext Cache (which is also handled internally by CommManager), which significantly reduces connection time when reconnecting.

To be honest, I used Claude to help me throughout the process, guiding the AI toward the desired architecture and reviewing every proposed code change. Centralizing communication in a single class, using Kotlin Flows, and creating a state machine makes it more robust and easier to follow than before (at least for me).

I attached a document created by Claude that explains all the changes.
[consolidated_changes.md](https://github.com/user-attachments/files/25693911/consolidated_changes.md)

I know this is a huge refactor, and there are some risks involved in merging this.
I've been testing this for the last few days on my car and it feels stable (both via Wi-Fi with the Wi-Fi Companion app and via USB).
I will continue troubleshooting issues with the initial USB connection, which sometimes takes a long time (the same happened before, and now I feel it works a bit better). Before continuing, I wanted to share these changes.

Let me know your thoughts.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/utils/Settings.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/utils/Settings.kt
@@ -102,11 +102,11 @@ class Settings(context: Context) {
             prefs.edit().putStringSet("key-codes", list.toSet()).apply()
         }
 
-    var logVerbosity: LogExporter.Verbosity
-        get() = LogExporter.Verbosity.entries.getOrElse(prefs.getInt("log-verbosity", LogExporter.Verbosity.INFO.ordinal)) { LogExporter.Verbosity.INFO }
-        set(value) { prefs.edit().putInt("log-verbosity", value.ordinal).apply() }
+    var exporterLogLevel: LogExporter.LogLevel
+        get() = LogExporter.LogLevel.entries.getOrElse(prefs.getInt("log-level", LogExporter.LogLevel.INFO.ordinal)) { LogExporter.LogLevel.INFO }
+        set(value) { prefs.edit().putInt("log-level", value.ordinal).apply() }
 
-    val logLevel: Int get() = logVerbosity.logLevel
+    val logLevel: Int get() = exporterLogLevel.logLevel
 
     var viewMode: ViewMode
         get() {
--- a/app/src/main/java/com/andrerinas/headunitrevived/main/SettingsFragment.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/main/SettingsFragment.kt
@@ -26,7 +26,6 @@ import com.andrerinas.headunitrevived.main.settings.SettingsAdapter
 import com.andrerinas.headunitrevived.utils.Settings
 import com.andrerinas.headunitrevived.utils.LocaleHelper
 import com.andrerinas.headunitrevived.BuildConfig
-import java.util.Locale
 import com.google.android.material.appbar.MaterialToolbar
 import com.google.android.material.button.MaterialButton
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
@@ -826,6 +825,25 @@ class SettingsFragment : Fragment() {
         // --- Debug Settings ---
         items.add(SettingItem.CategoryHeader("debug", R.string.category_debug))
 
+        val logLevels = com.andrerinas.headunitrevived.utils.LogExporter.LogLevel.entries
+        val logLevelNames = logLevels.map { it.name.lowercase().replaceFirstChar { c -> c.uppercase() } }.toTypedArray()
+        items.add(SettingItem.SettingEntry(
+            stableId = "logLevel",
+            nameResId = R.string.log_level,
+            value = settings.exporterLogLevel.name.lowercase().replaceFirstChar { it.uppercase() },
+            onClick = {
+                val currentIndex = logLevels.indexOf(settings.exporterLogLevel)
+                MaterialAlertDialogBuilder(requireContext(), R.style.DarkAlertDialog)
+                    .setTitle(R.string.log_level)
+                    .setSingleChoiceItems(logLevelNames, currentIndex) { dialog, which ->
+                        settings.exporterLogLevel = logLevels[which]
+                        dialog.dismiss()
+                        updateSettingsList()
+                    }
+                    .show()
+            }
+        ))
+
         items.add(SettingItem.SettingEntry(
             stableId = "captureLog",
             nameResId = if (com.andrerinas.headunitrevived.utils.LogExporter.isCapturing) R.string.stop_log_capture else R.string.start_log_capture,
@@ -835,28 +853,9 @@ class SettingsFragment : Fragment() {
                 if (com.andrerinas.headunitrevived.utils.LogExporter.isCapturing) {
                     com.andrerinas.headunitrevived.utils.LogExporter.stopCapture()
                 } else {
-                    com.andrerinas.headunitrevived.utils.LogExporter.startCapture(context, settings.logVerbosity)
+                    com.andrerinas.headunitrevived.utils.LogExporter.startCapture(context, settings.exporterLogLevel)
                 }
                 updateSettingsList()
-            }
-        ))
-
-        val verbosityLevels = com.andrerinas.headunitrevived.utils.LogExporter.Verbosity.entries
-        val verbosityNames = verbosityLevels.map { it.name.lowercase().replaceFirstChar { c -> c.uppercase() } }.toTypedArray()
-        items.add(SettingItem.SettingEntry(
-            stableId = "logVerbosity",
-            nameResId = R.string.log_verbosity,
-            value = settings.logVerbosity.name.lowercase().replaceFirstChar { it.uppercase() },
-            onClick = {
-                val currentIndex = verbosityLevels.indexOf(settings.logVerbosity)
-                MaterialAlertDialogBuilder(requireContext(), R.style.DarkAlertDialog)
-                    .setTitle(R.string.log_verbosity)
-                    .setSingleChoiceItems(verbosityNames, currentIndex) { dialog, which ->
-                        settings.logVerbosity = verbosityLevels[which]
-                        dialog.dismiss()
-                        updateSettingsList()
-                    }
-                    .show()
             }
         ))
 
@@ -869,7 +868,7 @@ class SettingsFragment : Fragment() {
                 if (com.andrerinas.headunitrevived.utils.LogExporter.isCapturing) {
                     com.andrerinas.headunitrevived.utils.LogExporter.stopCapture()
                 }
-                val logFile = com.andrerinas.headunitrevived.utils.LogExporter.saveLogToPublicFile(context, settings.logVerbosity)
+                val logFile = com.andrerinas.headunitrevived.utils.LogExporter.saveLogToPublicFile(context, settings.exporterLogLevel)
                 updateSettingsList()
 
                 if (logFile != null) {
--- a/app/src/main/java/com/andrerinas/headunitrevived/utils/LogExporter.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/utils/LogExporter.kt
@@ -13,7 +13,7 @@ import java.util.Locale
 
 object LogExporter {
 
-    enum class Verbosity(val filter: String, val logLevel: Int) {
+    enum class LogLevel(val filter: String, val logLevel: Int) {
         VERBOSE("*:V", Log.VERBOSE),
         DEBUG("*:D", Log.DEBUG),
         INFO("*:I", Log.INFO),
@@ -56,7 +56,7 @@ object LogExporter {
      * Unlike [saveLogToPublicFile], this captures everything from the moment it is called,
      * bypassing the small shared ring buffer.
      */
-    fun startCapture(context: Context, verbosity: Verbosity) {
+    fun startCapture(context: Context, verbosity: LogLevel) {
         stopCapture()
         val logDir = context.getExternalFilesDir(null) ?: return
         logDir.mkdirs()
@@ -100,7 +100,7 @@ object LogExporter {
      *   copies its content into a fresh export file so the original capture file is preserved.
      * - Otherwise: dumps the current logcat ring buffer.
      */
-    fun saveLogToPublicFile(context: Context, verbosity: Verbosity): File? {
+    fun saveLogToPublicFile(context: Context, verbosity: LogLevel): File? {
         val logDir = context.getExternalFilesDir(null) ?: return null
         if (!logDir.exists()) logDir.mkdirs()
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -78,11 +78,11 @@
     <string name="back">Back</string>
     <string name="debug_mode">Debug Mode</string>
     <string name="debug_mode_description">Toggle more Logging</string>
-    <string name="log_verbosity">Log Verbosity</string>
-    <string name="log_verbosity_description">Filter log level when exporting</string>
-    <string name="start_log_capture">Start Log Capture</string>
+    <string name="log_level">Log Level</string>
+    <string name="log_level_description">Filter log level when exporting</string>
+    <string name="start_log_capture">Start Capturing Logs</string>
     <string name="start_log_capture_description">Begin recording full logs for debugging</string>
-    <string name="stop_log_capture">Stop Log Capture</string>
+    <string name="stop_log_capture">Stop Capturing Logs</string>
     <string name="stop_log_capture_description">Recording in progress… tap to stop</string>
     <string name="export_logs">Export Logs</string>
     <string name="export_logs_description">Save logs for debugging</string>


andreknieriem:
Wow, that sounds great. But I've just merged some new features and addition from me and @andrecuellar and I think, that this will conflict. Maybe you pull the latest main branch again und merge it and me and @andrecuellar will review it, so it does not break the new features! Anyways: A big thank you!!!

nicoruy:
Ok, let me do the merging. I will need some time to review the code for the new features and merge. I may do that during the weekend.

nicoruy:
Log files were truncated on low-resource devices. I added a settings option to start and stop logging, enabling export of full logs. I also added an option to set the level (replacing the debugMode setting) that works for filtering both runtime and exported logs.

nicoruy:
After analyzing the logs, I identified the cause of the connection failure when connecting via USB. A deadlock occurred because the same lock simultaneously blocked both read and write operations on the port. By narrowing the lock's scope to only cover state changes (connect, disconnect, or reset), the deadlock was eliminated. 
The connection now functions smoothly, and Android Auto projection activates within just a few seconds.

andreknieriem:
Sounds really great! I am very excited when the merge can happen :)

andrecuellar:
Hey @nicoruy! Just went through the whole PR — really nice work 👏

The CommManager approach with the state machine is way cleaner than what we had with broadcast intents flying around everywhere. And the two-phase transport startup makes a lot of sense, I always wondered why we sometimes missed the first frame.

But honestly, the thing I'm most excited about is the USB deadlock fix. I've been dealing with USB connection issues for a while now (see #147 and my PRs #139 and #160). The main problem is with wireless Android Auto USB dongles — they change their device ID on every reconnect and re-enumerate (connect/disconnect) 3-5 times before settling. I built an auto-connect feature and a stability check to handle that, but even with those workarounds the auto-connect only works about 50% of the time. I've been exporting and analyzing logs trying to figure out why, and I ended up shipping it as "experimental" because I couldn't pin down the root cause. If the deadlock in `UsbAccessoryConnection` was causing the handshake to stall during those re-enumeration cycles, that would explain everything. Really curious to test this on my head unit.

Funny thing — for some strange reason, it works much better for me if I just enable "Auto-connect USB" and leave the experimental "USB stability check" disabled. Sometimes it still doesn't connect, but more often than not it does. With the stability check enabled it's actually less reliable, which never made sense to me. Maybe the deadlock was causing timeouts during the stability wait window too. I kept the stability check in the settings though, because I believe it works correctly for some people — so it might still be useful depending on the device.

About the conflicts — yeah, quite a bit landed in main recently from my two PRs and some additions from @andreknieriem. Here's what will need attention during the merge:

- **`AapService.kt`** — this will be the biggest one. I added: auto-reconnect after USB disconnect (re-plugging the phone auto-reconnects without user interaction), retry logic with a configurable max attempts slider, USB adapter stability check (waits for a wireless dongle to stop re-enumerating before connecting), auto-start on USB (brings the app to foreground when a USB device is plugged in), and race condition guards using an `isConnecting` AtomicBoolean on the companion object to prevent duplicate connection attempts from UsbAttachedActivity/AapService/HomeFragment firing simultaneously. Since your rewrite is the bigger change, I'd suggest starting from your version and carefully pulling in those features
- **`HomeFragment.kt`** — I added a configurable auto-connect priority system. There's now an `AutoConnectFragment` with a drag-to-reorder list where users can set the execution order of: reconnect last session, self mode, and single USB auto-connect. HomeFragment iterates `settings.autoConnectPriorityOrder` instead of using hardcoded priorities. Your version will need to incorporate this loop and the single USB auto-connect method
- **`AapProjectionActivity.kt`** — I added a retry button that shows after 10s on the loading screen when it gets stuck, with a dialog offering "Retry Android Auto" (soft reconnect) or "Reset USB Connection" (full USB reset with stability-aware polling, respecting the user's configured stability timeout). That will need to hook into your connection state flow instead of `AapService.isConnected`
- **`SettingsFragment.kt`** — new toggles for auto-connect single USB device, USB adapter stability check (with an experimental warning dialog that reverts the toggle on cancel), auto-start on USB, a max retry attempts slider, and a link to the new `AutoConnectFragment` for managing auto-connect priorities
- **`Settings.kt`** — new properties: `autoConnectSingleUsbDevice`, `usbStabilityCheck`, `usbStabilityTimeout` (3-50s slider), `maxAutoRetryAttempts` (1-10 slider), `autoConnectPriorityOrder` (persisted ordered list), and constants `AUTO_CONNECT_LAST_SESSION`, `AUTO_CONNECT_SELF_MODE`, `AUTO_CONNECT_SINGLE_USB`
- **New files that need CommManager wiring** — `AutoConnectFragment.kt` (reorderable priority screen with drag handles and up/down arrows), `AutomationActivity.kt`, `AutoConnectAdapter.kt` + `AutoConnectTouchCallback` (drag-and-drop RecyclerView adapter). These are standalone so they should mostly just work, but any references to `AapService.isConnected` will need to switch to `commManager.isConnected`
- **`strings.xml` + all locale files** — new strings for retry button, auto-connect, stability check, experimental warnings, auto-start on USB, etc. across all 11 locales (including a new Arabic locale `values-ar/`)

I think `git merge main` into your branch is the way to go (rebasing 19 commits would be painful). For the heavy conflicts in AapService and HomeFragment, I'd keep your version as the base and re-add the features from main on top.

Happy to help review once you push the merge — I can check that the auto-connect, stability check, and retry features still work correctly with CommManager. If your deadlock fix improves the reliability of my auto-connect feature, that would be a huge win for everyone using wireless dongles.

Looking forward to testing this! 🚀

nicoruy:
Hey @andrecuellar, thanks for taking the time to review the PR! 
I also struggled with the connection issue for a while, trying to find its cause. I thought it was related to the first frame missing, which, based on the app's behavior, seemed likely. After spending some time tracing the connection process and investigating the missing first frame, I decided to perform that refactor. I'm glad you liked the new approach!

During the process, I implemented a way to cache part of the SSL handshake, which significantly improves reconnection times. I also found that the issue was not related to the frame missing because it only occurred on the first connection, when the full handshake was performed. Unfortunately, logs were trimmed on my car headunit (Android 4.4). That's why I added options to start and stop log recording as a way to save longer logs to the log file. With full logs, I discovered that during the initial handshake, there was a moment when the phone didn't respond and it timed out. The reason was that the same lock was used for both sending and receiving data. The discovery response was never delivered to the phone because the send pool was locked, and communication was stalled until timeout. 
Adding the new log entries to the settings screen introduces a few more strings. How are you managing translations for those new texts?

I'm working on the merge and doing some additional testing. The app is working flawlessly over USB now, but I'm having some issues over WiFi that I didn't notice before. It has more lag and is sometimes unstable. It might be related to my car headunit. WiFi has always been slow and buggy in my car. I don't think I changed anything specific related to data transfer (changes for preventing the deadlock were made on UsbAccesoryConnection), so that doesn't seem to be the cause. Are you experiencing those WiFi issues with my code, or is it working the same as before?

Also, since the connection now happens in less than 10 seconds, I wasn't able to see or test the retry button. If this continues, we can consider removing it because it would add unnecessary complexity. The same applies to the number of retries. What do you think?

I'm considering buying one of those AA WiFi dongles to connect to AA over a 5 GHz WiFi instead of using the one from my car. It might be a good opportunity to test this feature as well. Any recommendations?

If I don't find any major issues, I think I will be able to have everything merged by the weekend, maybe sooner.
Your explanations about your implementation will help me a lot with the process. Thanks!

mrkontrast-coder:
I have the same situation, my Android Head Unit only has 2.4 Ghz WiFi, so I'm using AndroidAuto Dongle (Carlinkit Mini Ultra, witch has 5 GHz), and it works better then build in WiFi.

andrecuellar:
Hey @nicoruy, replying to your questions:

### Translations

I add the translations using AI for all locale files. The available locales are auto-detected at build time by scanning `values-*/strings.xml` directories and stored in `BuildConfig.AVAILABLE_LOCALES`, so you just need to add your new strings (like the log level/capture ones) to each file. I can handle adding the translations for your new strings once you're ready for the merge.

### Retry Button

I agree — since your deadlock fix resolves the root cause of the stuck loading screen, the retry button adds unnecessary complexity. Let's remove it. When you merge `main`, you can discard the retry button changes from #160:
- `retryButton`, `retryTimerRunnable`, `showRetryDialog()`, and `RETRY_DELAY_MS` in `AapProjectionActivity.kt`
- `retryConnection()`, `resetUsbAndReconnect()`, `ACTION_RETRY_CONNECTION`, and `ACTION_RESET_USB` in `AapService.kt`
- The `maxAutoRetryAttempts` setting and slider in `Settings.kt` and `SettingsFragment.kt`
- The retry-related strings: `retry`, `retry_connection_title`, `retry_connection_option`, `reset_usb_option`, `reset_usb_timeout`, `max_auto_retry_attempts`, `max_auto_retry_attempts_description`
- The layout changes to `activity_headunit.xml` (keep the original loading overlay structure without the retry button)

### WiFi Stability

The WiFi instability might be a separate issue — I've seen similar reports in #161. Let's focus on getting USB solid first with your deadlock fix, then investigate WiFi separately.

andrecuellar:
## Code Review & Merge Conflict Resolution Guide

I've done a thorough review of your changes. Below are merge conflict resolution suggestions for integrating the features from #139 (auto-connect single USB device) and notes on what to discard from #160 (retry button). I've also included some code improvement observations.

---

## Part 1: Merge Conflict Resolution (by file)

### `Settings.kt`

When merging `main`, add these settings (from #139). Do **NOT** include `maxAutoRetryAttempts` (that's from the retry button in #160, which we're removing).

```kotlin
var autoConnectSingleUsbDevice: Boolean
    get() = prefs.getBoolean("auto-connect-single-usb", false)
    set(value) { prefs.edit().putBoolean("auto-connect-single-usb", value).apply() }

var usbStabilityCheck: Boolean
    get() = prefs.getBoolean("usb-stability-check", false)
    set(value) { prefs.edit().putBoolean("usb-stability-check", value).apply() }

var usbStabilityTimeout: Int
    get() = prefs.getInt("usb-stability-timeout", 10)
    set(value) { prefs.edit().putInt("usb-stability-timeout", value).apply() }

var autoStartOnUsb: Boolean
    get() = prefs.getBoolean("auto-start-on-usb", false)
    set(value) { prefs.edit().putBoolean("auto-start-on-usb", value).apply() }

var autoConnectPriorityOrder: List<String>
    get() {
        val stored = prefs.getString("auto-connect-priority-order", null)
        val order = if (stored.isNullOrEmpty()) {
            DEFAULT_AUTO_CONNECT_ORDER.toMutableList()
        } else {
            stored.split(",").toMutableList()
        }
        // Append any missing methods (forward compat)
        for (method in DEFAULT_AUTO_CONNECT_ORDER) {
            if (method !in order) order.add(method)
        }
        order.retainAll(DEFAULT_AUTO_CONNECT_ORDER)
        return order
    }
    set(value) {
        prefs.edit().putString("auto-connect-priority-order", value.joinToString(",")).apply()
    }
```

And add these constants to the `companion object`:

```kotlin
const val AUTO_CONNECT_LAST_SESSION = "last-session"
const val AUTO_CONNECT_SELF_MODE = "self-mode"
const val AUTO_CONNECT_SINGLE_USB = "single-usb"

val DEFAULT_AUTO_CONNECT_ORDER = listOf(
    AUTO_CONNECT_LAST_SESSION,
    AUTO_CONNECT_SELF_MODE,
    AUTO_CONNECT_SINGLE_USB
)
```

---

### `HomeFragment.kt`

Replace the hardcoded auto-connect priorities with the configurable priority loop from #139. This needs to be adapted to use your Flow-based state management (replace `AapService.isConnected` checks with `commManager.isConnected`):

```kotlin
// Execute auto-connect methods in user-defined priority order
for (methodId in appSettings.autoConnectPriorityOrder) {
    if (commManager.isConnected) break
    when (methodId) {
        Settings.AUTO_CONNECT_LAST_SESSION -> {
            if (appSettings.autoConnectLastSession && !hasAttemptedAutoConnect) {
                hasAttemptedAutoConnect = true
                attemptAutoConnect()
            }
        }
        Settings.AUTO_CONNECT_SELF_MODE -> {
            if (appSettings.autoStartSelfMode && !hasAutoStarted) {
                hasAutoStarted = true
                startSelfMode()
            }
        }
        Settings.AUTO_CONNECT_SINGLE_USB -> {
            if (appSettings.autoConnectSingleUsbDevice && !hasAttemptedSingleUsbAutoConnect) {
                hasAttemptedSingleUsbAutoConnect = true
                attemptSingleUsbAutoConnect()
            }
        }
    }
}
```

Also add the missing field and method:

```kotlin
private var hasAttemptedSingleUsbAutoConnect = false

private fun attemptSingleUsbAutoConnect() {
    val appSettings = App.provide(requireContext()).settings
    if (!appSettings.autoConnectSingleUsbDevice ||
        !appSettings.hasAcceptedDisclaimer ||
        commManager.isConnected) return

    AppLog.i("HomeFragment: Requesting single-USB auto-connect via AapService")
    ContextCompat.startForegroundService(requireContext(),
        Intent(requireContext(), AapService::class.java).apply {
            action = AapService.ACTION_CHECK_USB
        })
}
```

---

### `AapService.kt`

This is the biggest conflict area. Key changes to integrate from #139:

**1. Add `ACTION_CHECK_USB` handler** in `onStartCommand` — should call `performSingleUsbConnect()` when invoked.

**2. Add single-USB auto-connect methods** (adapted to use `commManager`):

```kotlin
private var usbStabilityJob: Job? = null

private fun performSingleUsbConnect() {
    val settings = App.provide(this).settings
    if (!settings.autoConnectSingleUsbDevice) return
    if (commManager.isConnected ||
        commManager.connectionState.value is CommManager.ConnectionState.Connecting) return

    val usbManager = getSystemService(Context.USB_SERVICE) as UsbManager
    val devices = usbManager.deviceList.values.toList()

    if (devices.size != 1) {
        AppLog.i("Single-USB auto-connect: ${devices.size} devices found, need exactly 1")
        return
    }

    val device = devices.first()
    val deviceCompat = UsbDeviceCompat(device)

    if (UsbDeviceCompat.isInAccessoryMode(device)) {
        AppLog.i("Single-USB auto-connect: already in accessory mode")
        serviceScope.launch { commManager.connect(device) }
        return
    }

    if (!usbManager.hasPermission(device)) {
        AppLog.i("Single USB auto-connect: device found but no permission")
        return
    }

    if (settings.usbStabilityCheck) {
        startSingleUsbStabilityCheck(device, deviceCompat, usbManager)
    } else {
        AppLog.i("Single-USB auto-connect: switching to accessory mode")
        serviceScope.launch(Dispatchers.IO) {
            UsbAccessoryMode(usbManager).connectAndSwitch(device)
        }
    }
}

private fun startSingleUsbStabilityCheck(device: UsbDevice, deviceCompat: UsbDeviceCompat, usbManager: UsbManager) {
    val timeout = App.provide(this).settings.usbStabilityTimeout
    AppLog.i("USB stability: Starting ${timeout}s timer for ${deviceCompat.uniqueName}")
    usbStabilityJob?.cancel()
    usbStabilityJob = serviceScope.launch {
        delay(timeout * 1000L)
        // Re-check device is still present
        val stillPresent = usbManager.deviceList.values.any {
            UsbDeviceCompat(it).uniqueName == deviceCompat.uniqueName
        }
        if (stillPresent) {
            AppLog.i("USB stability: Device ${deviceCompat.uniqueName} stable after ${timeout}s, connecting")
            withContext(Dispatchers.IO) {
                UsbAccessoryMode(usbManager).connectAndSwitch(device)
            }
        }
    }
}

private fun cancelUsbStabilityCheck() {
    usbStabilityJob?.cancel()
    usbStabilityJob = null
}
```

**3. Expand `onUsbAttach`** to handle single-USB auto-connect:

```kotlin
override fun onUsbAttach(device: UsbDevice) {
    if (UsbDeviceCompat.isInAccessoryMode(device)) {
        AppLog.i("USB accessory device attached, connecting.")
        checkAlreadyConnectedUsb(force = true)
    } else if (App.provide(this).settings.autoConnectSingleUsbDevice) {
        AppLog.i("Normal USB device attached, checking single-USB auto-connect")
        performSingleUsbConnect()
    }
}
```

**4. Add cleanup calls** — `cancelUsbStabilityCheck()` in both `onDestroy()` and `onDisconnected()`.

**5. Discard from #160** (retry button):
- `ACTION_RETRY_CONNECTION` and `ACTION_RESET_USB` constants and handlers
- `retryConnection()` and `resetUsbAndReconnect()` methods
- `maxAutoRetryAttempts` usage

---

### `AapProjectionActivity.kt`

**Discard ALL retry button code from #160.** Your version is the desired state here. When you see conflict markers, choose YOUR version and do NOT include:
- `private var retryButton: Button?`
- `private val retryTimerRunnable`
- `retryButton` references in `onCreate`, `onPause`, `onResume`
- `showRetryDialog()` method
- `RETRY_DELAY_MS` constant
- The restructured `loading_overlay` layout with retry button in `activity_headunit.xml`

Your two-phase handshake + the deadlock fix resolve the stuck loading issue.

---

### `UsbAttachedActivity.kt`

Add the `autoStartOnUsb` feature from #139 (adapted to use `commManager`):

```kotlin
val settings = Settings(this)
if (settings.autoStartOnUsb && !App.provide(this).commManager.isConnected) {
    AppLog.i("USB auto-start: launching app")
    try {
        startActivity(Intent(this, MainActivity::class.java).apply {
            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
        })
    } catch (e: Exception) {
        AppLog.w("Could not start UI from USB auto-start: ${e.message}")
    }
}
```

---

### `SettingsFragment.kt`

When merging, add the following pending variables from #139:

```kotlin
private var pendingAutoStartOnUsb: Boolean? = null
private var pendingUsbStabilityCheck: Boolean? = null
private var pendingUsbStabilityTimeout: Int? = null
```

Add UI entries in `updateSettingsList()` for:
- "Auto-start on USB" toggle
- "Auto-connect Settings" navigation entry (opens `AutoConnectFragment`)
- "USB adapter stability check" toggle (with experimental feature warning dialog)
- "Stability timeout" slider (1-30s)

Add Bluetooth permission launcher (`bluetoothPermissionLauncher`) for runtime `BLUETOOTH_CONNECT` permission on Android 12+.

Do **NOT** include `pendingMaxAutoRetryAttempts` or the retry attempts slider (from #160).

---

### New files from #139 (keep as-is)

These files are new and won't conflict — just keep them:
- `AutoConnectFragment.kt` + `fragment_auto_connect.xml`
- `AutoConnectAdapter.kt` + `item_auto_connect.xml`
- `layout_setting_item_slider.xml`
- `ic_arrow_down.xml`, `ic_arrow_up.xml`, `ic_drag_handle.xml`
- Navigation entry in `nav_graph.xml`

---

### `strings.xml` (all locales)

Keep all strings from #139 (auto-connect, stability, BT permission). Discard these from #160:
- `retry`, `retry_connection_title`, `retry_connection_option`
- `reset_usb_option`, `reset_usb_timeout`
- `max_auto_retry_attempts`, `max_auto_retry_attempts_description`

Your new log level/capture strings should be added to all 12 locale files. I can help with translations once you're ready.

---

## Part 2: Code Improvements & Bug Fixes

### Bug: `enableEdgeToEdge()` crash on API < 21

**File:** `AapProjectionActivity.kt`

The minSdk is 16 (Android 4.1). `enableEdgeToEdge()` requires API 21+. Current `main` wraps this in a version check that your version dropped:

```kotlin
// Your code (crashes on Android 4.1-4.4):
enableEdgeToEdge()

// Should be:
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
    enableEdgeToEdge()
}
```

---

### Bug: `destroy()` cancels scope right after launching cleanup

**File:** `CommManager.kt`

```kotlin
fun destroy() {
    _scope.launch {
        withContext(Dispatchers.IO) {
            doDisconnect()
        }
    }
    _scope.cancel()  // <-- cancels the just-launched coroutine!
}
```

`_scope.cancel()` cancels all child coroutines, including the one just launched. `doDisconnect()` may never execute, leaving the transport and connection alive. Fix:

```kotlin
fun destroy() {
    doDisconnect()  // synchronous, safe from onDestroy
    _scope.cancel()
}
```

Or consider removing `destroy()` entirely and just using `disconnect()` in `onDestroy()`, since `CommManager` lives as a singleton in `AppComponent` and its scope should live for the app lifetime.

---

### Dead code: `catch` blocks in `send()` methods never execute

**File:** `CommManager.kt`

```kotlin
fun send(keyCode: Int, isPress: Boolean) {
    if (_connectionState.value is ConnectionState.TransportStarted) {
        try {
            _scope.launch {
                _transport?.send(keyCode, isPress)
            }
        } catch (e: Exception) {
            _scope.launch { ... disconnect() }
        }
    }
}
```

`_scope.launch {}` does not throw exceptions from the coroutine body back to the caller — it returns a `Job`. The `catch` block is dead code. Since `onQuit` already handles disconnects when the transport fails, simplify to:

```kotlin
fun send(keyCode: Int, isPress: Boolean) {
    if (_connectionState.value is ConnectionState.TransportStarted) {
        _scope.launch { _transport?.send(keyCode, isPress) }
    }
}
```

---

### Race: TOCTOU in `connect()` guard

**File:** `CommManager.kt`

All three `connect()` overloads check `_connectionState.value is Connecting` then proceed. Two concurrent callers could both pass the guard. Consider using a `Mutex`:

```kotlin
private val _connectMutex = kotlinx.coroutines.sync.Mutex()

suspend fun connect(device: UsbDevice) = withContext(Dispatchers.IO) {
    _connectMutex.withLock {
        if (_connectionState.value is ConnectionState.Connecting) return@withContext
        // ... rest of connect logic
    }
}
```

This mostly works today because the UI serializes connect calls, but if USB attach + WiFi discovery fire simultaneously, you could hit this.

---

### Fragility: `drop(1)` in disconnect observer

**File:** `AapService.kt`

```kotlin
commManager.connectionState
    .filterIsInstance<CommManager.ConnectionState.Disconnected>()
    .drop(1) // skip the StateFlow's initial Disconnected emission
    .collect { state -> onDisconnected(state) }
```

`drop(1)` assumes the first emission is always the initial `Disconnected`. If a fast connect/disconnect cycle happens before the coroutine subscribes, you'll miss a real disconnect. Safer:

```kotlin
commManager.connectionState
    .filterIsInstance<CommManager.ConnectionState.Disconnected>()
    .collect { state ->
        if (hasEverConnected) onDisconnected(state)
    }
```

Where `hasEverConnected` is set to `true` in `onConnected()`.

---

### Ordering: `startForegroundService` called from IO dispatcher

**File:** `NetworkListFragment.kt`

```kotlin
lifecycleScope.launch(Dispatchers.IO) {
    commManager.connect(socket)
    ContextCompat.startForegroundService(ctx, ...) // <-- called from IO thread
}
```

`startForegroundService` must run on the main thread and within the 5-second `startForeground()` deadline. Calling it from `Dispatchers.IO` after a slow `connect()` may violate the deadline. Start the foreground service first:

```kotlin
ContextCompat.startForegroundService(ctx, Intent(ctx, AapService::class.java).apply {
    action = AapService.ACTION_CONNECT_SOCKET
})
lifecycleScope.launch(Dispatchers.IO) {
    commManager.connect(socket)
}
```

---

### Unbounded `pendingTlsData` growth

**File:** `AapSslContext.kt`

The BUFFER_UNDERFLOW handler keeps appending: `pendingTlsData += buffer.copyOfRange(6, size)`. On a misbehaving device, this could grow until OOM (the 15s timeout is the only safety net). Consider adding a size cap:

```kotlin
pendingTlsData += buffer.copyOfRange(6, size)
if (pendingTlsData.size > 65536) {
    AppLog.e("SSL Handshake: accumulated TLS data exceeds 64KB, aborting")
    return false
}
```

---

### Missing pre-KitKat fullscreen flag

**File:** `AapProjectionActivity.kt`

Current `main` has a pre-KitKat fullscreen workaround that your version dropped:

```kotlin
if (Build.VERSION.SDK_INT < Build.VERSION_CODES.KITKAT && settings.startInFullscreenMode) {
    window.addFlags(android.view.WindowManager.LayoutParams.FLAG_FULLSCREEN)
}
```

Without this, devices running Android 4.1-4.3 won't enter fullscreen mode since `SystemUI.apply` alone doesn't achieve true fullscreen on API < 19.

---

### `isConnected` doesn't include `Connecting` state

**File:** `CommManager.kt`

Several callers use `!commManager.isConnected` to decide whether to start a new connection. If the state is `Connecting`, these callers might start a second attempt. Consider either adding `isConnecting` as a separate convenience property, or documenting that callers who want to prevent duplicate connections should also check for `Connecting`.

---

## Part 3: Positive Feedback

**Deadlock fix** — Excellent root cause analysis. Moving `bulkTransfer()` outside `synchronized` blocks and calling `close()` before acquiring `sStateLock` in `disconnect()` is the correct fix. The `@Volatile` annotations ensure thread visibility. Minor note: `releaseInterface()` will always fail after `close()` — you could skip it to avoid the noisy log.

**SSL session caching** — Great design. Using a consistent `("android-auto", 5277)` peer key in `createSSLEngine()` for JSSE session lookup, and keeping a singleton `AapSslContext` across transport recreations, is clean and effective.

**Two-phase transport** — Splitting `start()` into `startHandshake()` and `startReading()` is a solid architectural improvement that eliminates the race between surface creation and message pumping.

**CommManager state machine** — The sealed class state machine with Flow-based observation is much cleaner than the broadcast-based approach. Good use of `_disconnectJob?.join()` for sequentializing disconnect/reconnect cycles.

andrecuellar:
Hey @nicoruy, I've submitted a code review with inline suggestions to make resolving the merge conflicts with `main` less painful. The review covers:

**Code improvements (8 inline comments):**
1. **CommManager.kt** — `destroy()` race condition (bug), dead `catch` in `send()` methods, `connect()` TOCTOU race, `isConnected` clarification
2. **AapService.kt** — `drop(1)` fragility in disconnect observer
3. **AapSslContext.kt** — Unbounded `pendingTlsData` growth
4. **NetworkListFragment.kt** — `startForegroundService` called from IO dispatcher
5. **UsbAccessoryConnection.kt** — Deadlock fix positive feedback + `releaseInterface` note

**Merge conflict guidance (6 inline comments):**
1. **AapService.kt** — `onUsbAttach` expansion, `ACTION_CHECK_USB` handler, `checkAlreadyConnectedUsb` fallback
2. **HomeFragment.kt** — Priority-based auto-connect loop
3. **Settings.kt** — New settings properties
4. **UsbAttachedActivity.kt** — Auto-start on USB feature

**In the review body:**
- Bugs outside diff hunks (`enableEdgeToEdge` crash, pre-KitKat fullscreen)
- Retry button removal checklist (#160)
- SettingsFragment merge notes
- New files from #139 (no conflicts)
- Strings guidance

For the merge, I recommend `git merge main`. Let me know if you need help with anything!

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
```diff
@@ -0,0 +1,452 @@
...
+     * After [destroy], the CommManager instance must not be used again.
+     */
+    fun destroy() {
+        _scope.launch {
+            withContext(Dispatchers.IO) {
+                doDisconnect()
+            }
+        }
+        _scope.cancel()
+    }
```
andrecuellar, line 451: **Bug: `destroy()` cancels scope right after launching cleanup coroutine**

`_scope.cancel()` on line 450 immediately cancels all children, including the coroutine just launched on line 445. `doDisconnect()` may never execute, leaving transport and connection alive.

```suggestion
    fun destroy() {
        doDisconnect()
        _scope.cancel()
    }
```

`doDisconnect()` is not a suspend function, so calling it directly is safe. Alternatively, consider removing `destroy()` entirely and just using `disconnect()` in `AapService.onDestroy()`, since `CommManager` is a singleton in `AppComponent` with app-lifetime scope.

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
```diff
@@ -0,0 +1,452 @@
...
+                _scope.launch {
+                    _transport?.send(keyCode, isPress)
+                }
+            } catch (e: Exception) {
+                _scope.launch {
+                    _connectionState.emit(ConnectionState.Error("Send failed"))
+                    disconnect()
+                }
+            }
+        }
```
andrecuellar, line 349: **Dead code: `catch` block never executes**

`_scope.launch {}` does not throw exceptions from the coroutine body — it returns a `Job`. The `catch` on line 343 is dead code. Same pattern applies to the other two `send()` overloads below.

Since `onQuit` already handles disconnects when transport fails, simplify:

```suggestion
    fun send(keyCode: Int, isPress: Boolean) {
        if (_connectionState.value is ConnectionState.TransportStarted) {
            _scope.launch { _transport?.send(keyCode, isPress) }
        }
    }
```

Apply the same simplification to `send(data: ByteArray, ...)` and `send(sensor: SensorEvent.SensorBatch)` below.

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
```diff
@@ -0,0 +1,452 @@
...
+
+    /**
+     * Opens a USB accessory connection to [device].
+     *
+     * On success emits [ConnectionState.Connected] and persists the device as the last-used
+     * connection so it can be auto-reconnected on the next launch.
+     */
+    suspend fun connect(device: UsbDevice) = withContext(Dispatchers.IO) {
+        // Another caller already started the connection — do nothing.
+        if (_connectionState.value is ConnectionState.Connecting)
```
andrecuellar, line 156: **Race condition: TOCTOU in `Connecting` guard**

Two concurrent callers on `Dispatchers.IO` could both read `Disconnected`, both pass this guard, and both proceed to disconnect + reconnect. Consider using a `Mutex`:

```kotlin
private val _connectMutex = kotlinx.coroutines.sync.Mutex()

suspend fun connect(device: UsbDevice) = withContext(Dispatchers.IO) {
    _connectMutex.withLock {
        if (_connectionState.value is ConnectionState.Connecting) return@withContext
        // ... rest of connect logic
    }
}
```

Same pattern applies to all three `connect()` overloads. This mostly works today because the UI serializes connect calls, but if USB attach + WiFi discovery fire simultaneously, you could hit this.

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
```diff
@@ -0,0 +1,452 @@
...
+
+    /** Public read-only view of [_connectionState]. */
+    val connectionState = _connectionState.asStateFlow()
+
+    /**
+     * `true` while a physical connection exists, regardless of whether the AAP transport
+     * handshake has completed. Covers [ConnectionState.Connected], [ConnectionState.StartingTransport],
+     * and [ConnectionState.TransportStarted].
+     */
+    val isConnected: Boolean
```
andrecuellar, line 129: **Observation: `isConnected` doesn't include `Connecting` state**

Several callers use `!commManager.isConnected` to decide whether to start a new connection. If the state is `Connecting`, these callers might start a second attempt. Consider adding:

```kotlin
val isConnecting: Boolean
    get() = connectionState.value is ConnectionState.Connecting
```

Or document that callers should also check `Connecting` to prevent duplicate connections.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
```diff
@@ -73,297 +86,391 @@ class AapService : Service(), UsbReceiver.Listener {
...
-            startWirelessServer();
+    /**
+     * Observes [CommManager.ConnectionState.Disconnected] and calls onDisconnect
+     * `drop(1)` skips the initial `Disconnected` value that `StateFlow` replays on subscribe.
+     */
+    private fun observeDisconnectedState() {
+        serviceScope.launch {
+            commManager.connectionState
+                .filterIsInstance<CommManager.ConnectionState.Disconnected>()
+                .drop(1) // skip the StateFlow's initial Disconnected emission on startup
```
andrecuellar, line 162: **Fragility: `drop(1)` assumes initial emission timing**

If a fast connect/disconnect cycle happens before this coroutine subscribes, you'll miss a real disconnect. Safer pattern:

```kotlin
commManager.connectionState
    .filterIsInstance<CommManager.ConnectionState.Disconnected>()
    .collect { state ->
        if (hasEverConnected) onDisconnected(state)
    }
```

Where `hasEverConnected` is set to `true` in the `TransportStarted` observer.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapSslContext.kt
```diff
@@ -21,26 +21,70 @@ class AapSslContext(keyManager: SingleKeyKeyManager): AapSsl {
...
+                        SSLEngineResult.Status.BUFFER_UNDERFLOW -> {
+                            // Partial TLS record — read the next USB packet and append it.
+                            // The outer loop re-enters NEED_UNWRAP and retries unwrap with
+                            // the now-larger buffer (pendingTlsData non-empty → no extra USB read).
+                            val size = connection.recvBlocking(buffer, buffer.size, 2000, false)
+                            if (size <= 6) {
+                                AppLog.e("SSL Handshake: Receive failed during accumulation ($size)")
+                                return false
+                            }
+                            pendingTlsData += buffer.copyOfRange(6, size)
```
andrecuellar, line 74: **Safety: unbounded `pendingTlsData` growth**

On a misbehaving device that never produces a complete TLS record, this array grows until OOM. The 15s timeout is the only safety net. Consider a size cap:

```kotlin
pendingTlsData += buffer.copyOfRange(6, size)
if (pendingTlsData.size > 65536) {
    AppLog.e("SSL Handshake: accumulated TLS data exceeds 64KB, aborting")
    return false
}
```

app/src/main/java/com/andrerinas/headunitrevived/main/NetworkListFragment.kt
```diff
@@ -165,12 +170,16 @@ class NetworkListFragment : Fragment(), NetworkDiscovery.Listener {
...
-                // If we have a socket, we need to pass it to AapService for reuse!
-                if (socket != null && socket.isConnected) {
-                    AapService.pendingSocket = socket
+                val ctx = context ?: return@runOnUiThread
+                lifecycleScope.launch(Dispatchers.IO) {
+                    if (socket != null && socket.isConnected)
+                        App.provide(ctx).commManager.connect(socket)
+                    else
+                        App.provide(ctx).commManager.connect(ip, 5277)
+                    ContextCompat.startForegroundService(ctx, Intent(ctx, AapService::class.java).apply {
```
andrecuellar, line 179: **Issue: `startForegroundService` called from `Dispatchers.IO`**

This is inside a `lifecycleScope.launch(Dispatchers.IO)` block. `startForegroundService` must run on the main thread and the system enforces a 5-second `startForeground()` deadline. Calling it after a potentially slow `commManager.connect()` may violate the deadline.

Start the foreground service first (on main thread), then connect:

```kotlin
ContextCompat.startForegroundService(ctx, Intent(ctx, AapService::class.java).apply {
    action = AapService.ACTION_CONNECT_SOCKET
})
lifecycleScope.launch(Dispatchers.IO) {
    if (socket != null && socket.isConnected)
        App.provide(ctx).commManager.connect(socket)
    else
        App.provide(ctx).commManager.connect(ip, 5277)
}
```

Same issue at line ~320 where `CoroutineScope(Dispatchers.Main).launch` is used instead of `lifecycleScope` (potential leak if fragment is destroyed).

app/src/main/java/com/andrerinas/headunitrevived/connection/UsbAccessoryConnection.kt
```diff
@@ -150,26 +149,29 @@ class UsbAccessoryConnection(private val usbMgr: UsbManager, private val device:
...
-            endpointOut = null                                               // Output EP
+            endpointIn = null
+            endpointOut = null
 
             if (usbDeviceConnection != null) {
                 var bret = false
                 if (usbInterface != null) {
-                    bret = usbDeviceConnection!!.releaseInterface(usbInterface)
+                    // releaseInterface() may fail since close() was already called; log and continue.
+                    bret = try { usbDeviceConnection!!.releaseInterface(usbInterface) } catch (_: Exception) { false }
```
andrecuellar, line 168: **Deadlock fix: excellent work + minor note**

The fix is correct and elegant — calling `close()` before `sStateLock` on line 155 immediately aborts in-flight `bulkTransfer()` calls, breaking the deadlock. The `@Volatile` annotations ensure thread visibility. This directly resolves the root cause of the stuck loading screen reported in multiple issues.

Minor: since `close()` was already called on line 155, `releaseInterface()` here will always throw (connection is closed). The try-catch handles it, but you could skip the call to avoid the noisy "Error releaseInterface()" log on every disconnect.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
```diff
@@ -73,297 +86,391 @@ class AapService : Service(), UsbReceiver.Listener {
...
-        accessoryConnection = connectionFactory(intent, this);
-        
-        if (accessoryConnection == null) {
-            AppLog.e("Cannot create connection from intent");
-            return;
+    // -------------------------------------------------------------------------
+    // USB
+    // -------------------------------------------------------------------------
+
+    override fun onUsbAttach(device: UsbDevice) {
```
andrecuellar, line 319: **Merge conflict: Expand for single-USB auto-connect (#139)**

Current `main` adds single-USB auto-connect handling for non-accessory devices. When merging, expand:

```kotlin
override fun onUsbAttach(device: UsbDevice) {
    if (UsbDeviceCompat.isInAccessoryMode(device)) {
        AppLog.i("USB accessory device attached, connecting.")
        checkAlreadyConnectedUsb(force = true)
    } else if (App.provide(this).settings.autoConnectSingleUsbDevice) {
        AppLog.i("Normal USB device attached, checking single-USB auto-connect")
        performSingleUsbConnect()
    }
}
```

`performSingleUsbConnect()` auto-connects when exactly one USB device is present — handles dongles that change USB ID on every reconnect (#147). Add this method along with `startSingleUsbStabilityCheck()` and `cancelUsbStabilityCheck()` from #139, adapted to use `commManager.connect(device)` instead of direct transport calls.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
```diff
@@ -73,297 +86,391 @@ class AapService : Service(), UsbReceiver.Listener {
...
+                if (commManager.isConnected) commManager.disconnect()
             }
-            ACTION_CHECK_USB -> {
-                checkAlreadyConnectedUsb();
+            ACTION_CONNECT_SOCKET        -> {
+                // Caller already invoked commManager.connect(socket); the connectionState
+                // observer in observeConnectedState() handles the rest — nothing to do here.
             }
-            else -> {
+            ACTION_CHECK_USB             -> checkAlreadyConnectedUsb(force = true)
```
andrecuellar, line 305: **Merge conflict: Add `performSingleUsbConnect()` fallback**

Current `main` also triggers `performSingleUsbConnect()` as a fallback here when `checkAlreadyConnectedUsb()` doesn't find an accessory-mode device. Add the following methods from #139 (adapted to CommManager):

- `performSingleUsbConnect()` — checks for exactly 1 USB device, handles permission, stability check, AOA switch
- `startSingleUsbStabilityCheck()` — configurable timer for dongles that re-enumerate
- `cancelUsbStabilityCheck()` — cancels stability timer (call in `onDestroy` and `onDisconnected`)
- Field: `private var usbStabilityJob: Job? = null`

Do **NOT** add `retryConnection()`, `resetUsbAndReconnect()`, `ACTION_RETRY_CONNECTION`, or `ACTION_RESET_USB` — those are from the retry button (#160) which we're removing.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
```diff
@@ -73,297 +86,391 @@ class AapService : Service(), UsbReceiver.Listener {
...
+    /**
+     * Scans currently connected USB devices and connects to any that are already in
+     * Android Open Accessory (AOA) mode, or attempts to switch a known device into AOA mode.
+     *
+     * @param force When `true`, bypasses the [autoConnectLastSession] guard. Use `true` when
+     *              called in response to an actual USB attach event or from [UsbAttachedActivity],
+     *              because the user has explicitly plugged in a device. Use `false` (default)
+     *              for the startup scan in [onCreate].
+     */
+    private fun checkAlreadyConnectedUsb(force: Boolean = false) {
```
andrecuellar, line 349: **Merge conflict: Add single-USB fallback path (#139)**

Current `main` adds a fallback after the existing last-session loop: if `autoConnectSingleUsbDevice` is enabled and exactly one device is present, connect to it regardless of the allowed list. Add after the existing loop:

```kotlin
// Fallback: single USB auto-connect (any single device)
if (settings.autoConnectSingleUsbDevice) {
    performSingleUsbConnect()
}
```

app/src/main/java/com/andrerinas/headunitrevived/main/HomeFragment.kt
```diff
@@ -99,24 +73,35 @@ class HomeFragment : Fragment() {
...
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
+                AapService.scanningState.collect { updateWifiButtonFeedback(it) }
+            }
+        }
+
         val appSettings = App.provide(requireContext()).settings
 
         // 1. Priority: Auto-Connect last session (WiFi/USB)
-        if (appSettings.autoConnectLastSession && !hasAttemptedAutoConnect && !AapService.isConnected) {
+        if (appSettings.autoConnectLastSession && !hasAttemptedAutoConnect && !commManager.isConnected) {
```
andrecuellar, line 91: **Merge conflict: Replace hardcoded priorities with configurable loop (#139)**

Current `main` replaces the hardcoded 1/2 priority with a user-configurable loop. Adapt to use `commManager.isConnected`:

```kotlin
for (methodId in appSettings.autoConnectPriorityOrder) {
    if (commManager.isConnected) break
    when (methodId) {
        Settings.AUTO_CONNECT_LAST_SESSION -> {
            if (appSettings.autoConnectLastSession && !hasAttemptedAutoConnect) {
                hasAttemptedAutoConnect = true
                attemptAutoConnect()
            }
        }
        Settings.AUTO_CONNECT_SELF_MODE -> {
            if (appSettings.autoStartSelfMode && !hasAutoStarted) {
                hasAutoStarted = true
                startSelfMode()
            }
        }
        Settings.AUTO_CONNECT_SINGLE_USB -> {
            if (appSettings.autoConnectSingleUsbDevice && !hasAttemptedSingleUsbAutoConnect) {
                hasAttemptedSingleUsbAutoConnect = true
                attemptSingleUsbAutoConnect()
            }
        }
    }
}
```

Also add `hasAttemptedSingleUsbAutoConnect` field and `attemptSingleUsbAutoConnect()` method that sends `ACTION_CHECK_USB` to `AapService`.

app/src/main/java/com/andrerinas/headunitrevived/utils/Settings.kt
```diff
@@ -102,11 +102,11 @@ class Settings(context: Context) {
             prefs.edit().putStringSet("key-codes", list.toSet()).apply()
         }
 
-    var debugMode: Boolean
-        get() = prefs.getBoolean("debug-mode", false)
-        set(value) {
-            prefs.edit().putBoolean("debug-mode", value).apply()
-        }
+    var exporterLogLevel: LogExporter.LogLevel
```
andrecuellar, line 105: **Merge conflict: Add settings from #139**

When merging, add these settings nearby (do NOT include `maxAutoRetryAttempts` — retry button removed):

```kotlin
var autoConnectSingleUsbDevice: Boolean
    get() = prefs.getBoolean("auto-connect-single-usb", false)
    set(value) { prefs.edit().putBoolean("auto-connect-single-usb", value).apply() }

var usbStabilityCheck: Boolean
    get() = prefs.getBoolean("usb-stability-check", false)
    set(value) { prefs.edit().putBoolean("usb-stability-check", value).apply() }

var usbStabilityTimeout: Int
    get() = prefs.getInt("usb-stability-timeout", 10)
    set(value) { prefs.edit().putInt("usb-stability-timeout", value).apply() }

var autoStartOnUsb: Boolean
    get() = prefs.getBoolean("auto-start-on-usb", false)
    set(value) { prefs.edit().putBoolean("auto-start-on-usb", value).apply() }

var autoConnectPriorityOrder: List<String>
    get() { /* priority ordering with migration safety */ }
    set(value) { prefs.edit().putString("auto-connect-priority-order", value.joinToString(",")).apply() }
```

Plus companion constants: `AUTO_CONNECT_LAST_SESSION`, `AUTO_CONNECT_SELF_MODE`, `AUTO_CONNECT_SINGLE_USB`, `DEFAULT_AUTO_CONNECT_ORDER`.

app/src/main/java/com/andrerinas/headunitrevived/app/UsbAttachedActivity.kt
```diff
@@ -36,15 +37,17 @@ class UsbAttachedActivity : Activity() {
             return
         }
 
-        if (App.provide(this).transport.isAlive) {
+        if (App.provide(this).commManager.connectionState.value is CommManager.ConnectionState.TransportStarted) {
```
andrecuellar, line 40: **Merge conflict: Add `autoStartOnUsb` feature (#139)**

Current `main` adds an auto-start check before this line. When a USB device is attached and `autoStartOnUsb` is enabled, launch `MainActivity`:

```kotlin
val settings = Settings(this)
if (settings.autoStartOnUsb && !App.provide(this).commManager.isConnected) {
    AppLog.i("USB auto-start: launching app")
    try {
        startActivity(Intent(this, MainActivity::class.java).apply {
            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
        })
    } catch (e: Exception) {
        AppLog.w("Could not start UI from USB auto-start: ${e.message}")
    }
}
```

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
```diff
@@ -0,0 +1,452 @@
...
+        get() = connectionState.value.let {
+            it is ConnectionState.Connected ||
+            it is ConnectionState.StartingTransport ||
+            it is ConnectionState.HandshakeComplete ||
+            it is ConnectionState.TransportStarted
+        }
```
andrecuellar: **Observation: `isConnected` doesn't include `Connecting` state**

Several callers use `!commManager.isConnected` to decide whether to start a new connection. If the state is `Connecting`, these callers might start a second attempt. Consider adding:

```kotlin
val isConnecting: Boolean
    get() = connectionState.value is ConnectionState.Connecting
```

Or document that callers should also check `Connecting` to prevent duplicate connections.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
andrecuellar: **Fragility: `drop(1)` assumes initial emission timing**

If a fast connect/disconnect cycle happens before this coroutine subscribes, you'll miss a real disconnect. Safer pattern:

```kotlin
commManager.connectionState
    .filterIsInstance<CommManager.ConnectionState.Disconnected>()
    .collect { state ->
        if (hasEverConnected) onDisconnected(state)
    }
```

Where `hasEverConnected` is set to `true` in the `TransportStarted` observer.

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
andrecuellar, line 451: **Bug: `destroy()` cancels scope right after launching cleanup coroutine**

`_scope.cancel()` on line 450 immediately cancels all children, including the coroutine just launched on line 445. `doDisconnect()` may never execute, leaving transport and connection alive.

Fix: call `doDisconnect()` directly — it is not a suspend function:

```kotlin
fun destroy() {
    doDisconnect()
    _scope.cancel()
}
```

Alternatively, consider removing `destroy()` entirely and just using `disconnect()` in `AapService.onDestroy()`, since `CommManager` is a singleton in `AppComponent` with app-lifetime scope.

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
```diff
@@ -0,0 +1,452 @@
...
+    // -----------------------------------------------------------------------------------------
+
+    /** Sends a key press or release event to the phone. */
+    fun send(keyCode: Int, isPress: Boolean) {
+        if (_connectionState.value is ConnectionState.TransportStarted) {
+            try {
+                _scope.launch {
+                    _transport?.send(keyCode, isPress)
+                }
+            } catch (e: Exception) {
```
andrecuellar, line 343: **Dead code: `catch` block never executes**

`_scope.launch {}` returns a `Job` — exceptions from the launched coroutine body are propagated to the `SupervisorJob`, not thrown to the caller. This `catch` is dead code.

Same pattern applies to the other two `send()` overloads below. Since `onQuit` already handles disconnects when transport fails, simplify all three to:

```kotlin
fun send(keyCode: Int, isPress: Boolean) {
    if (_connectionState.value is ConnectionState.TransportStarted) {
        _scope.launch { _transport?.send(keyCode, isPress) }
    }
}
```

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
```diff
@@ -0,0 +1,452 @@
...
+            if (_connection?.connect() ?: false) {
+                settings.saveLastConnection(type = Settings.CONNECTION_TYPE_USB, usbDevice = UsbDeviceCompat.getUniqueName(device))
+                _connectionState.emit(ConnectionState.Connected)
+            } else {
+                _connectionState.emit(ConnectionState.Disconnected())
+            }
+        } catch (e: Exception) {
+            _connectionState.emit(ConnectionState.Error("Connection failed: ${e.message}"))
+            disconnect()
+        }
```
andrecuellar, line 178: **Race condition: TOCTOU in `Connecting` guard**

Two concurrent callers on `Dispatchers.IO` could both read `Disconnected`, both pass this guard, and both proceed to disconnect + reconnect. Consider using a `Mutex`:

```kotlin
private val _connectMutex = kotlinx.coroutines.sync.Mutex()

suspend fun connect(device: UsbDevice) = withContext(Dispatchers.IO) {
    _connectMutex.withLock {
        if (_connectionState.value is ConnectionState.Connecting) return@withContext
        // ... rest of connect logic
    }
}
```

Same pattern applies to all three `connect()` overloads. This mostly works today because the UI serializes connect calls, but if USB attach + WiFi discovery fire simultaneously, you could hit this.

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
andrecuellar, line 135: **Observation: `isConnected` doesn't include `Connecting` state**

Several callers use `!commManager.isConnected` to decide whether to start a new connection. If the state is `Connecting`, these callers might start a second attempt. Consider adding:

```kotlin
val isConnecting: Boolean
    get() = connectionState.value is ConnectionState.Connecting
```

Or document that callers should also check `Connecting` to prevent duplicate connections.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
```diff
@@ -73,297 +86,391 @@ class AapService : Service(), UsbReceiver.Listener {
...
+    /**
+     * Observes [CommManager.ConnectionState.Disconnected] and calls onDisconnect
+     * `drop(1)` skips the initial `Disconnected` value that `StateFlow` replays on subscribe.
+     */
+    private fun observeDisconnectedState() {
+        serviceScope.launch {
+            commManager.connectionState
+                .filterIsInstance<CommManager.ConnectionState.Disconnected>()
+                .drop(1) // skip the StateFlow's initial Disconnected emission on startup
+                .collect { state -> onDisconnected(state) }
```
andrecuellar, line 163: **Fragility: `drop(1)` assumes initial emission timing**

If a fast connect/disconnect cycle happens before this coroutine subscribes, you'll miss a real disconnect. Safer pattern:

```kotlin
commManager.connectionState
    .filterIsInstance<CommManager.ConnectionState.Disconnected>()
    .collect { state ->
        if (hasEverConnected) onDisconnected(state)
    }
```

Where `hasEverConnected` is set to `true` in the `TransportStarted` observer.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapSslContext.kt
```diff
@@ -21,26 +21,70 @@ class AapSslContext(keyManager: SingleKeyKeyManager): AapSsl {
...
+                            AppLog.d("SSL Handshake: partial TLS record, ${pendingTlsData.size} B buffered")
```
andrecuellar, line 75: **Safety: unbounded `pendingTlsData` growth**

On a misbehaving device that never produces a complete TLS record, this array grows until OOM. The 15s timeout is the only safety net. Consider adding a size cap:

```kotlin
pendingTlsData += buffer.copyOfRange(6, size)
if (pendingTlsData.size > 65536) {
    AppLog.e("SSL Handshake: accumulated TLS data exceeds 64KB, aborting")
    return false
}
```

app/src/main/java/com/andrerinas/headunitrevived/main/NetworkListFragment.kt
```diff
@@ -165,12 +170,16 @@ class NetworkListFragment : Fragment(), NetworkDiscovery.Listener {
...
+                        action = AapService.ACTION_CONNECT_SOCKET
```
andrecuellar, line 180: **Issue: `startForegroundService` called from `Dispatchers.IO`**

This is inside a `lifecycleScope.launch(Dispatchers.IO)` block. `startForegroundService` must run on the main thread and the system enforces a 5-second `startForeground()` deadline. Calling it after a potentially slow `commManager.connect()` may violate the deadline.

Start the foreground service first (on main thread), then connect:

```kotlin
ContextCompat.startForegroundService(ctx, Intent(ctx, AapService::class.java).apply {
    action = AapService.ACTION_CONNECT_SOCKET
})
lifecycleScope.launch(Dispatchers.IO) {
    if (socket != null && socket.isConnected)
        App.provide(ctx).commManager.connect(socket)
    else
        App.provide(ctx).commManager.connect(ip, 5277)
}
```

Same issue at line ~323 where `CoroutineScope(Dispatchers.Main).launch` is used instead of `lifecycleScope` (potential leak if fragment is destroyed).

app/src/main/java/com/andrerinas/headunitrevived/connection/UsbAccessoryConnection.kt
```diff
@@ -150,26 +149,29 @@ class UsbAccessoryConnection(private val usbMgr: UsbManager, private val device:
         }
     }
 
-    override fun disconnect() {                                           // Release interface and close USB device connection. Called only by usb_disconnect()
-        synchronized(sLock) {
+    override fun disconnect() {
+        // close() is thread-safe and immediately aborts any in-flight bulkTransfer(),
+        // so both sendBlocking and recvBlocking unblock within milliseconds.
+        usbDeviceConnection?.close()
+
```
andrecuellar, line 156: **Deadlock fix: excellent work + minor note**

The fix is correct and elegant — calling `close()` before `sStateLock` immediately aborts in-flight `bulkTransfer()` calls, breaking the deadlock. The `@Volatile` annotations ensure thread visibility. This directly resolves the root cause of the stuck loading screen reported in multiple issues.

Minor: since `close()` was already called on line 155, `releaseInterface()` here will always throw (connection is closed). The try-catch handles it, but you could skip the call to avoid the noisy "Error releaseInterface()" log on every disconnect.

app/src/main/java/com/andrerinas/headunitrevived/main/NetworkListFragment.kt
```diff
@@ -307,7 +316,11 @@ class NetworkListFragment : Fragment(), NetworkDiscovery.Listener {
 
         override fun onClick(v: View) {
             if (v.id == android.R.id.button2) {
-                ContextCompat.startForegroundService(context, AapService.createIntent(v.getTag(R.integer.key_data) as String, context))
+                val ip = v.getTag(R.integer.key_data) as String
+                ContextCompat.startForegroundService(context, Intent(context, AapService::class.java).apply {
+                    action = AapService.ACTION_CONNECT_SOCKET
+                })
+                CoroutineScope(Dispatchers.Main).launch { App.provide(context).commManager.connect(ip, 5277) }
```
andrecuellar, line 323: **Leak: `CoroutineScope(Dispatchers.Main)` creates an unscoped coroutine**

This coroutine is not tied to any lifecycle. If the fragment is destroyed during the connect call, the coroutine keeps running and holds a reference to the destroyed fragment's `context`.

Since `AddressAdapter` is a `private class` (not inner), it can't access `viewLifecycleOwner` directly. Fix: pass the fragment's `lifecycleScope` via the constructor:

```kotlin
private class AddressAdapter(
    private val context: Context,
    private val fragmentManager: FragmentManager,
    private val scope: CoroutineScope
)

// Then use:
scope.launch(Dispatchers.IO) { App.provide(context).commManager.connect(ip, 5277) }
```

Also needs `Dispatchers.IO` since `connect()` does blocking socket I/O.

andrecuellar:
@nicoruy I've created a draft PR with the merge conflicts resolved and compile errors fixed, if it helps as a guide: #180

It compiles successfully but I haven't tested it on a device yet, so proceed with caution — use it as a reference, not as a drop-in.

I have compiled the .apk, I will test it in a few hours

mrkontrast-coder:
Can you provide the apk, for testing ?

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
nicoruy: The goal of isConnected is to determine whether the full connection process is complete (the state in which we can start the transport handshake).
I think it would be better to keep it as is to avoid confusion and check for ConnectionState.Connecting in addition to isConnected when needed. We might need to review all the places where AapServices.isConnected to see if it makes sense. There might be a more elegant solution than checking the connection every time.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
nicoruy: I don't think adding another variable to the AapService to set the connection status is the best approach. The responsibility for the connection status should be handled by CommManager. The reason 'Disconnected' is triggered the first time is that connectionState is initialized to that value.

`private val _connectionState = MutableStateFlow<ConnectionState>(ConnectionState.Disconnected())`

I agree that using drop isn't the best approach. I think we can create a new Undetermined connectionState and initialize the Flow with that state.

`private val _connectionState = MutableStateFlow<ConnectionState>(ConnectionState.Undetermined())`

and then remove the drop filter.

app/src/main/java/com/andrerinas/headunitrevived/main/NetworkListFragment.kt
nicoruy: Agree with the change. Also I found that ACTION_CONNECT_SOCKET might not be needed because that intent action is not used anymore. We can remove the action constant and the parameter to start the service.

app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
nicoruy: We can add this additional attrinute but it's also possible to check for the ConnectionState for this specific value.

app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
nicoruy: Same as previous comment. We can add the Undetermiend state and remove drop

nicoruy:
Hey, thank you for working on merging my code! 
I was struggling with the entire reconnection intents and methods. Removing the retry buttons and logic simplifies things a lot. 
I added comments above specific changes that I think should be handled differently. Please take a look at them during the merge process. I also noticed that part of the code in main involved updating the MediaSessionState based on a callback from transport. This was something new.
As part of the merge, I moved that to commManager because Transport is private. 

The startReading method would look like this:

`
suspend fun startReading() = withContext(Dispatchers.IO) {
        if (_connectionState.value !is ConnectionState.HandshakeComplete) return@withContext

        try {
            _mediaSession = MediaSessionCompat(context, "HeadunitRevived").apply { isActive = true }
            _transport?.onAudioFocusStateChanged = { isPlaying -> updateMediaSessionState(isPlaying) }

            _transport?.aapAudio?.requestFocusChange(
                AudioManager.STREAM_MUSIC,
                AudioManager.AUDIOFOCUS_GAIN,
                AudioManager.OnAudioFocusChangeListener { }
            )
            _transport?.startReading()
            _connectionState.emit(ConnectionState.TransportStarted)
        } catch (e: Exception) {
            _connectionState.emit(ConnectionState.Error("Start reading failed: ${e.message}"))
            disconnect()
        }
    }
`

updateMediaSessionState method is the one currently on AapService.

mediaSession is released on disconnect (and on destroy)

```
    fun disconnect(sendByeBye: Boolean = true) {
        if (_connectionState.value is ConnectionState.Disconnected) return

        _mediaSession?.isActive = false
        _mediaSession?.release()
        _mediaSession = null

        _connectionState.value = ConnectionState.Disconnected()
        _disconnectJob = _scope.launch { doDisconnect(sendByeBye) }
    }
```

Since you've already removed the retry logic, I will let you continue with the merge process. Please consider the items I mentioned above, and let me know when you have a version ready for review and test.

Thanks !!!

andrecuellar:
Hey @nicoruy! 👋

Just wanted to share some findings after testing on a low-spec head unit (Chinese Android unit with MediaTek SoC). The CommManager architecture is a great improvement in terms of code clarity and separation of concerns, but I think there might be a few performance-sensitive spots worth looking into — especially for devices with limited CPU cores.

Here's what I noticed:

### 1. `send()` wraps every message in a coroutine launch

In `CommManager.send()`, each call does `_scope.launch { _transport?.send(message) }`. Since the scope uses `Dispatchers.IO`, every send does two thread hops: caller → IO pool → sendHandler. Touch events alone fire at ~60Hz, so that's 60+ coroutine allocations per second (each one creates a Job + Continuation object), plus double the context switches compared to calling `transport.send()` directly (which just uses Android's zero-allocation `Handler.obtainMessage()` pool).

On a 2-4 core MediaTek device, the extra thread scheduling and GC pressure from all those tiny coroutine objects could definitely be noticeable.

### 2. Multiple concurrent StateFlow observers

Between AapService (4 observers), AapProjectionActivity (2), and HomeFragment (2), there are ~8 concurrent `filterIsInstance<>()` pipelines on the same `connectionState` flow. Each state transition gets processed by all of them. The original architecture used simple BroadcastReceivers which skip non-matching intents with zero overhead.

### 3. Two-phase handshake/reading adds video startup latency

The split between `startHandshake()` and `startReading()` (coordinated via StateFlow from AapProjectionActivity) adds a main-thread message queue delay between handshake completion and the read loop starting. The original `transport.start()` did both in one synchronous call, so frames started arriving immediately.

I think these are fixable without throwing away the CommManager — for example, the `send()` method could post directly to the transport's handler instead of launching a coroutine, and the handshake/reading phases could be combined. Would love to collaborate on this if you're interested!

andrecuellar:
## Performance improvement suggestions for CommManager

Following up on the lag investigation and the revert of PR #181 — the lag turned out to be from the tablet/WiFi itself, not CommManager. That said, here are some targeted improvements that would reduce overhead before the next merge attempt:

### 1. Simplify `send()` methods — remove unnecessary `_scope.launch {}`

`AapTransport.send()` is already thread-safe (it posts to a HandlerThread). The `_scope.launch {}` wrapper adds coroutine scheduling overhead on every message — this adds up on low-spec devices sending touch events at 60Hz.

**Before:**
```kotlin
fun send(message: AapMessage) {
    if (_connectionState.value is ConnectionState.TransportStarted) {
        try {
            _scope.launch { _transport?.send(message) }
        } catch (e: Exception) {
            _scope.launch {
                _connectionState.emit(ConnectionState.Error("Send failed"))
                disconnect()
            }
        }
    }
}
```

**After:**
```kotlin
fun send(message: AapMessage) {
    if (_connectionState.value is ConnectionState.TransportStarted) {
        _transport?.send(message)
    }
}
```

The `catch` block is dead code — `_scope.launch {}` returns a `Job` and never throws the coroutine body's exceptions to the caller. The `onQuit` callback already handles transport failures. Apply to all three `send()` overloads.

### 2. Fix `destroy()` race condition

As discussed in the inline review — `doDisconnect()` is not a suspend function, so call it directly:

```kotlin
fun destroy() {
    doDisconnect()
    _scope.cancel()
}
```

### 3. Consolidate StateFlow observers

Four separate `serviceScope.launch` blocks each create their own coroutine to observe the same `connectionState` flow with `filterIsInstance`. This means 4 active coroutines constantly collecting. Combine into one:

**Before (4 separate observers):**
```kotlin
private fun observeConnectedState() {
    serviceScope.launch {
        commManager.connectionState.filterIsInstance<ConnectionState.Connected>().collect { onConnected() }
    }
}
private fun observeTransportStartedState() { /* ... */ }
private fun observeErrorState() { /* ... */ }
private fun observeDisconnectedState() { /* ... */ }
```

**After (single observer):**
```kotlin
private fun observeConnectionState() {
    serviceScope.launch {
        commManager.connectionState.collect { state ->
            when (state) {
                is ConnectionState.Connected -> onConnected()
                is ConnectionState.TransportStarted -> {
                    accessoryHandshakeFailures = 0
                    sendBroadcast(Intent(ACTION_REQUEST_NIGHT_MODE_UPDATE).apply { setPackage(packageName) })
                }
                is ConnectionState.Error -> {
                    if (state.message.contains("Handshake failed")) onHandshakeFailed()
                }
                is ConnectionState.Disconnected -> {
                    if (hasEverConnected) onDisconnected(state)
                }
                else -> {}
            }
        }
    }
}
```

This also naturally addresses the `drop(1)` fragility — using `hasEverConnected` (set in `TransportStarted`) is more robust, as you suggested with the `Undetermined` initial state idea.

### 4. Re: `Undetermined` initial state

Agree with your suggestion to add `ConnectionState.Undetermined` as the initial state instead of using `drop(1)`. That's the cleanest solution. The single-observer pattern above would then just add `is ConnectionState.Undetermined -> {}` to the `when` block.

Let me know if you'd like me to submit these as a follow-up PR once you're ready for the next merge attempt.

Issue #173, opened 2026-03-03
USB autoconnect and auto launch feature not working


brownah:
I have found latest version with USB autoconnect does not work

Same as before need to manually select USB and then tap the device twice to grant usb permission then launch android auto. Every time you disconnect and reconnect the USB cable the device ID changes (/001/002/003 etc)

It does show dialog box to ask to auto launch app when USB device connected but even when ticked, does not do anything.

andrecuellar:
Hey @brownah, thanks for reporting this!

I'm the one who implemented the USB auto-connect feature. I've been testing it on my headunit, also with two different phones via USB OTG, and it's working for me — so I'd like to figure out what's different on your setup.

A few questions to help narrow this down:

1. **What device are you running the app on?** (head unit model, Android version)
2. **Are you connecting directly via USB cable or using a wireless Android Auto USB dongle?** If it's a dongle, which one?
3. **Which settings do you have enabled?** Could you check in Settings → Auto-Connect Settings and let me know:
   - Is "Auto-connect USB device" enabled?
   - What's the priority order shown?
   - Is "USB adapter stability check" enabled or disabled?
4. **Could you export the logs?** Go to Settings → Debug → Export Logs right after a failed auto-connect attempt. That would help a lot — the logs show exactly what the app is doing during the connection sequence and where it gets stuck
5. **About the permission dialog** — you mentioned it asks to auto-launch but doesn't do anything. Is this the Android system dialog that says "Open HeadUnit Revived when this USB device is connected"? After ticking that box, does it still require you to manually go into the USB list and tap the device?
6. **Could you check the app permissions?** Go to your device's Settings → Apps → HeadUnit Revived → Permissions and make sure the app has all necessary permissions, especially:
   - **Display over other apps** (overlay permission)
   - **Modify system settings**
   - Any other permissions the app requests

One thing I suspect — some head units block or reject the "Display over other apps" permission through their system apps. Without that permission, the app can't bring itself to the foreground automatically when a USB device is plugged in, which would explain exactly what you're seeing. If you could verify that permission is granted, that would help rule it out.

The changing device ID (`/001/002/003`) is a known issue with some dongles — the auto-connect feature was specifically designed to handle that by connecting to whatever single USB device is detected, regardless of its ID. So if it's not triggering at all, something else might be going on.

Any info you can share will help me track this down. Thanks!

brownah:
Hi @andrecuellar 
[...]

PR #181, opened 2026-03-05
Fix USB auto-connect failing when no permission is granted

andrecuellar:
## Summary
- Fix auto-connect silently failing when USB host permission is not granted (#173)
- Add `UsbManager.requestPermission()` calls to the auto-connect flow so the Android permission dialog appears
- Implement the empty `onUsbPermission()` callback in `AapService` to handle the result and connect on grant
- Add `resolveUsbDevice()` fallback in `UsbAttachedActivity` for head units that send `USB_DEVICE_ATTACHED` intents without the `UsbDevice` extra

> **Depends on #167 (CommManager refactor) being merged first.** This branch is based on the `guide/commmanager-merged` reference branch (#180).

## Root cause analysis (from #173 logs)

**Log 1 — Cold start:** App detects USB device `Google 18D1:4EE1`, logs `"Single USB auto-connect: device found but no permission"` **4 times** across HomeFragment, onCreate, and ACTION_CHECK_USB handlers, then gives up. Manual selection works fine.

**Log 2 — After disconnect:** Same pattern on reconnect, plus `UsbAttachedActivity.onCreate | No USB device` because the Xtrons Android 10 firmware sends `USB_DEVICE_ATTACHED` intents without the `UsbDevice` parcelable extra.

### `AapService.kt`
- **`requestUsbPermission(device)`** — new helper that creates a `PendingIntent` targeting `UsbReceiver.ACTION_USB_DEVICE_PERMISSION` and calls `UsbManager.requestPermission()`. Shows a "Requesting USB Permission…" toast (string already exists in all 12 locales)
- **`performSingleUsbConnect(device)`** — now calls `requestUsbPermission()` instead of silently giving up when `hasPermission()` returns false
- **`checkAlreadyConnectedUsb()`** — same fix for the last-session reconnect path: requests permission for known devices that lack it, then returns (the `onUsbPermission` callback handles the rest)
- **`onUsbPermission(granted, connect, device)`** — implemented (was an empty stub). On grant: connects directly if already in accessory mode, or switches to AOA mode via `connectAndSwitch()`. On deny: logs a warning

### `UsbAttachedActivity.kt`
- **`resolveUsbDevice(intent)`** — new helper that extracts `UsbDevice` from intent extras, falling back to `UsbManager.deviceList` when the extra is missing and exactly one device is connected. Used in both `onCreate()` and `onNewIntent()`

## Test plan
- [ ] Enable "Auto-connect USB device" → plug in one USB device without prior permission → Android USB permission dialog should appear → grant → auto-connect proceeds
- [ ] Deny USB permission → app logs warning, does not crash or retry endlessly
- [ ] On head units with empty USB intent extras (e.g. Xtrons) → `UsbAttachedActivity` falls back to `deviceList` instead of logging "No USB device" and exiting
- [ ] Multiple USB devices connected → `resolveUsbDevice()` returns null (doesn't pick one arbitrarily), existing behavior preserved
- [ ] Last-session USB reconnect without permission → permission dialog appears → grant → AOA mode switch proceeds
- [ ] Normal USB flow (manual selection from USB list) → no regression

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
@@ -364,7 +364,27 @@ class AapService : Service(), UsbReceiver.Listener {
         }
     }
 
-    override fun onUsbPermission(granted: Boolean, connect: Boolean, device: UsbDevice) {}
+    override fun onUsbPermission(granted: Boolean, connect: Boolean, device: UsbDevice) {
+        val deviceName = UsbDeviceCompat(device).uniqueName
+        if (granted) {
+            AppLog.i("USB permission granted for $deviceName")
+            if (UsbDeviceCompat.isInAccessoryMode(device)) {
+                serviceScope.launch { connectUsbWithRetry(device) }
+            } else {
+                val usbManager = getSystemService(Context.USB_SERVICE) as UsbManager
+                val usbMode = UsbAccessoryMode(usbManager)
+                serviceScope.launch(Dispatchers.IO) {
+                    if (usbMode.connectAndSwitch(device)) {
+                        AppLog.i("Successfully requested switch to accessory mode for $deviceName")
+                    } else {
+                        AppLog.w("USB permission granted but connectAndSwitch failed for $deviceName")
+                    }
+                }
+            }
+        } else {
+            AppLog.w("USB permission denied for $deviceName")
+        }
+    }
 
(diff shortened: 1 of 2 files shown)


andreknieriem:
@andrecuellar . Is here some code from the other PR from @nicoruy? And do you see the gemini review as well, or is it only me? I will merge it later. I think I get wifi direct mode work more automatic :) Need some tests.

PR #182, opened 2026-03-06
Revert "Fix USB auto-connect failing when no permission is granted"

andreknieriem:
Reverts andreknieriem/headunit-revived#181


andreknieriem:
It is running super laggy on my devices. I will revert it. @andrecuellar can you make your PR again. Sorry

andrecuellar:
> It is running super laggy on my devices. I will revert it. @andrecuellar can you make your PR again. Sorry

I'm so sorry! I'm going to see what happened

andreknieriem:
No everything is good. I think it was my tablet and my wifi were messy. But I was not able to reopen the PR. Can you please make a new one? :)

andrecuellar:
Hey @andreknieriem! No worries at all about the revert — totally understandable! 🙂

I did some digging into what might have caused the laggy behavior, and I think the root cause wasn't my USB permission fix (PR #181) itself, but rather the CommManager refactor (PR #167) that my PR was based on. Here's what I found:

### Most likely cause: coroutine-per-send overhead

`CommManager.send()` wraps every single message in `_scope.launch { ... }` on `Dispatchers.IO`. This means every touch event (~60 per second), every key press, every sensor update creates a new coroutine object (Job + Continuation) and does two thread hops (caller → IO pool → sendHandler) instead of one. The original `transport.send()` just calls `handler.sendMessage()` directly — zero allocation, single thread hop.

On a low-spec head unit with 2-4 CPU cores, those extra context switches and GC pressure from coroutine allocations add up fast.

### Other contributing factors

- **8 concurrent StateFlow observers** across AapService, AapProjectionActivity, and HomeFragment, all running `filterIsInstance<>()` on every state change
- **Two-phase video startup** (handshake + reading split, coordinated via StateFlow) adds latency vs. the original single `transport.start()` call
- **CommManager scope never gets cancelled** between connections — it accumulates coroutine jobs over the lifetime of the app

### What this means

My USB permission fix (the actual changes in PR #181 — `requestUsbPermission()`, `onUsbPermission()`, `resolveUsbDevice()`) shouldn't cause any performance issues on its own. The lag came from the underlying CommManager architecture. I'd be happy to re-submit my fix rebased on the current main (without CommManager) if that works for you!

Also, if we want to revisit the CommManager idea in the future, I think the performance issues are fixable — mainly by making `send()` post directly to the handler instead of launching coroutines, and combining the handshake/reading phases. Might be worth exploring down the road.

andreknieriem:
Yeah would be great without the new CommManager. And we should merge CommManager in the future. But I think it better to split the features ;)

Issue #137, opened 2026-02-18
In full-screen mode, make it possible to enable/disable only the navigation bar through the program settings.


BogenRu:

andreknieriem:
Can you please try https://github.com/andreknieriem/headunit-revived/releases/tag/v.2.0.0-beta

PR #184, opened 2026-03-06
Wifi direct

andreknieriem:
This will add better wifi direct functionality, especially for wireless helper 1.1.0 :) 
And it should fix: #137


tsabaia:
The Wi-Fi Direct(1.15.1) connection worked and was very stable, however, on my multimedia center, to start the Wi-Fi Direct connection, I need to go into the Android settings until I reach the Wi-Fi Direct search screen. After the devices connect, I just need to open HURev and Android Auto works. I have to repeat this process every time.

On the tablet, I only needed to access the settings once; the connection happens automatically afterward.

Will these changes planned for version 1.15.2 help with this type of problem?

PR #239, opened 2026-03-16
reverting the light theme, because it doesn't fit my asthetic view. B…

andreknieriem:
...because it doesn't fit my asthetic view. But if more people want a lighter theme, we can bring it back

### DIFF
--- a/app/src/main/res/drawable/bg.xml
+++ b/app/src/main/res/drawable/bg.xml
@@ -3,9 +3,9 @@
     android:shape="rectangle">
     <gradient
         android:type="radial"
-        android:centerX="0.5"
-        android:centerY="0.5"
-        android:gradientRadius="720"
-        android:startColor="#4A8FE7"
-        android:endColor="#F0F2F5" />
+        android:centerX="0.6"
+        android:centerY="0.6"
+        android:gradientRadius="620"
+        android:startColor="#296ddb"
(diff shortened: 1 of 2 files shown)


andrecuellar:
@andreknieriem 

It occurred to me that a switch called "use legacy background" or "use original background". Enabling it will use the bg.png image, while disabling it will cause the device to draw the background itself (which is what dark mode currently does).

This way, people who prefer it can continue using bg.png for dark/light mode.

What do you think?

andreknieriem:
Thought about that too, but I think there are soo many settings already. And I zhought about a BG select option too with the default bg.png.What do you think? So you can customize it.

andrecuellar:
> Thought about that too, but I think there are soo many settings already. And I zhought about a BG select option too with the default bg.png.What do you think? So you can customize it.

@andreknieriem 
I was thinking of saying something like, "Yes, I also thought about a background picker that uses bg.png by default," but I also thought that might add a lot of unnecessary complexity.

If more people want custom backgrounds later, we can expand it to an image picker (background selector).

A toggle sounds good to keep it simple, just 'Use gradient background' (off by default). When off, it uses the classic bg.png. When on, the app draws its own gradient (which already exists for dark mode). In this way, bg.png could also be used in dark mode, as it was originally used.

andreknieriem:
Sounds like a good idea.

PR #246, opened 2026-03-17
USB permission fix

Bastel2020:
My device changes its ID when it switches to Android Auto. This caused an access error. I fixed this issue in my previous pull request (navigation broadcasts), but the fix didn't make it into the release. I'm re-making the fix to match the current code.


andreknieriem:
Thank you!

PR #281, opened 2026-03-24
Kill app on disconnect feature

Tilak-03:
Hi! I'm not a developer but I'm a user of this app and wanted to add a feature that I felt was missing.

I used AI to help me write this code as I have no coding experience. So I'd really appreciate if you could review the code quality and suggest any improvements or changes needed.

The feature adds a toggle setting called "Kill app on disconnect" in the General settings section. When enabled, the app automatically closes itself 500ms after Android Auto projection disconnects. This works for all disconnect types — phone disconnecting, USB unplugging, or manual disconnect.

Changes made:
- Settings.kt: Added killOnDisconnect boolean preference
- CommManager.kt: Added kill logic in transportedQuited() and disconnect()
- SettingsFragment.kt: Added toggle UI in General settings
- strings.xml: Added string resources for the new setting

Happy to make any changes you suggest!

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/connection/CommManager.kt
@@ -341,6 +341,11 @@ class CommManager(
         _connectionState.value = ConnectionState.Disconnected(isClean, isUserExit = wasUserExit)
         // Transport already quit on its own — no ByeByeRequest needed (connection is dead).
         _disconnectJob = _scope.launch { doDisconnect(sendByeBye = false) }
+        if (settings.killOnDisconnect) {
+            android.os.Handler(android.os.Looper.getMainLooper()).postDelayed({
+                android.os.Process.killProcess(android.os.Process.myPid())
+            }, 500)
+        }
     }
 
     // -----------------------------------------------------------------------------------------
@@ -384,6 +389,11 @@ class CommManager(
 
         _connectionState.value = ConnectionState.Disconnected()
         _disconnectJob = _scope.launch { doDisconnect(sendByeBye) }
+        if (settings.killOnDisconnect) {
+            android.os.Handler(android.os.Looper.getMainLooper()).postDelayed({
+                android.os.Process.killProcess(android.os.Process.myPid())
+            }, 500)
+        }
     }
 
     /**
--- a/app/src/main/java/com/andrerinas/headunitrevived/main/SettingsFragment.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/main/SettingsFragment.kt
@@ -83,6 +83,8 @@ class SettingsFragment : Fragment() {
     
     // Flag to determine if the projection should stretch to fill the screen
     private var pendingStretchToFill: Boolean? = null
+
+    private var pendingKillOnDisconnect: Boolean? = null
     
     // Custom Insets
     private var pendingInsetLeft: Int? = null
@@ -135,6 +137,8 @@ class SettingsFragment : Fragment() {
         
         // Initialize local state for stretch to fill
         pendingStretchToFill = settings.stretchToFill
+
+        pendingKillOnDisconnect = settings.killOnDisconnect
         
         pendingInsetLeft = settings.insetLeft
         pendingInsetTop = settings.insetTop
@@ -259,6 +263,8 @@ class SettingsFragment : Fragment() {
 
         // Save the stretch to fill preference
         pendingStretchToFill?.let { settings.stretchToFill = it }
+
+        pendingKillOnDisconnect?.let { settings.killOnDisconnect = it }
         
         pendingInsetLeft?.let { settings.insetLeft = it }
         pendingInsetTop?.let { settings.insetTop = it }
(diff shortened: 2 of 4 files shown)


andreknieriem:
Thank you very much. I will add the missing labels and change the way the app exits, but thanks for your effort :)

andrecuellar:
@andreknieriem @Tilak-03

Personally, I believe that adding termination (or kill) logic could abruptly disrupt the lifecycle of several reconnection features, since I think there are more requests to automate connections, not disconnections (conflicts explained below)

This setting is a good idea, but it directly conflicts with several of my existing reconnection and auto-start features. An unsuspecting user could get stuck in an infinite loop (app kills itself, auto-connect relaunches it, disconnect triggers kill again), or lose the ability to recover from transient connection drops.

It's also worth noting that many headunits don't fully turn off — they stay powered or restart the app automatically. On those devices, having kill-on-disconnect active would completely prevent any auto-connect setting from working, since the app would kill itself before the reconnect logic has a chance to run. This could affect even more headunits than the ones we know about.

**Conflicting settings:**

- **Auto-connect last session** — On disconnect, `scheduleReconnectIfNeeded()` delays a few seconds then attempts to reconnect via USB or WiFi. The kill fires at 500ms, so the app dies before the reconnect can even start.
- **Auto-connect single USB device** — Same issue. USB auto-reconnect scheduling in `AapService` is killed before it executes.
- **Auto-start self mode** — If self mode is configured to auto-start on launch and kill-on-disconnect is enabled, the app could enter a launch → connect → disconnect → kill → relaunch cycle.
- **WiFi connection mode (Auto/Headunit Server)** — On unclean WiFi disconnects, the app waits 2 seconds then restarts WiFi discovery. Killed at 500ms before discovery restarts.
- **Wireless server mode** — When the head unit runs as a wireless server, the reconnect logic restarts the discovery loop after 2 seconds. Same problem.
- **Reconnecting overlay** — `AapProjectionActivity` has a watchdog that detects frame stalls and shows a "Connection interrupted" overlay, giving the user a chance to wait for recovery. With kill-on-disconnect, the user never sees the overlay — the app just dies on any transient drop.

**Additional concerns:**

- `Process.killProcess()` is a hard kill that doesn't run `onDestroy()`, so `doDisconnect()` (which sends ByeByeRequest to the phone) can be interrupted mid-cleanup, leaving the phone-side connection in a bad state.
- The kill doesn't distinguish between a clean disconnect (user unplugged) and a transient connection drop (brief signal loss). It should only trigger on clean or user-initiated disconnects.

I'd suggest showing a warning dialog when enabling this setting that lists the conflicting options and offers to disable them, or making kill-on-disconnect subordinate to auto-reconnect (i.e., only kill if no reconnect is scheduled).

Tilak-03:
> @andreknieriem @Tilak-03
> 
> Personally, I believe that adding termination (or kill) logic could abruptly disrupt the lifecycle of several reconnection features, since I think there are more requests to automate connections, not disconnections (conflicts explained below)
> 
> This setting is a good idea, but it directly conflicts with several of my existing reconnection and auto-start features. An unsuspecting user could get stuck in an infinite loop (app kills itself, auto-connect relaunches it, disconnect triggers kill again), or lose the ability to recover from transient connection drops.
> 
> It's also worth noting that many headunits don't fully turn off — they stay powered or restart the app automatically. On those devices, having kill-on-disconnect active would completely prevent any auto-connect setting from working, since the app would kill itself before the reconnect logic has a chance to run. This could affect even more headunits than the ones we know about.
> 
> **Conflicting settings:**
> 
> * **Auto-connect last session** — On disconnect, `scheduleReconnectIfNeeded()` delays a few seconds then attempts to reconnect via USB or WiFi. The kill fires at 500ms, so the app dies before the reconnect can even start.
> * **Auto-connect single USB device** — Same issue. USB auto-reconnect scheduling in `AapService` is killed before it executes.
> * **Auto-start self mode** — If self mode is configured to auto-start on launch and kill-on-disconnect is enabled, the app could enter a launch → connect → disconnect → kill → relaunch cycle.
> * **WiFi connection mode (Auto/Headunit Server)** — On unclean WiFi disconnects, the app waits 2 seconds then restarts WiFi discovery. Killed at 500ms before discovery restarts.
> * **Wireless server mode** — When the head unit runs as a wireless server, the reconnect logic restarts the discovery loop after 2 seconds. Same problem.
> * **Reconnecting overlay** — `AapProjectionActivity` has a watchdog that detects frame stalls and shows a "Connection interrupted" overlay, giving the user a chance to wait for recovery. With kill-on-disconnect, the user never sees the overlay — the app just dies on any transient drop.
> 
> **Additional concerns:**
> 
> * `Process.killProcess()` is a hard kill that doesn't run `onDestroy()`, so `doDisconnect()` (which sends ByeByeRequest to the phone) can be interrupted mid-cleanup, leaving the phone-side connection in a bad state.
> * The kill doesn't distinguish between a clean disconnect (user unplugged) and a transient connection drop (brief signal loss). It should only trigger on clean or user-initiated disconnects.
> 
> I'd suggest showing a warning dialog when enabling this setting that lists the conflicting options and offers to disable them, or making kill-on-disconnect subordinate to auto-reconnect (i.e., only kill if no reconnect is scheduled).

My bad I'm not a developer, my main issue is when I disconnect my tablet screen doesn't turn off automatically, it stays on and drains whole battery

andrecuellar:
@Tilak-03 

> My bad I'm not a developer, my main issue is when I disconnect my tablet screen doesn't turn off automatically, it stays on and drains whole battery

No worries, the ideas are very good; I will work to ensure they can be implemented without disrupting functions that could break down (The user will need to turn off the auto-connect and auto-start functionality if this option is enabled.)

Tilak-03:
> @Tilak-03
> 
> > My bad I'm not a developer, my main issue is when I disconnect my tablet screen doesn't turn off automatically, it stays on and drains whole battery
> 
> No worries, the ideas are very good; I will work to ensure they can be implemented without disrupting functions that could break down (The user will need to turn off the auto-connect and auto-start functionality if this option is enabled.)

Thank you so much for considering it :)

PR #282, opened 2026-03-24
Connection enhancements

andreknieriem:


andreknieriem:
Ok this breaks more, than it heals. Was worth a try. I will come back to this later.

PR #286, opened 2026-03-24
Safeguards for close-on-disconnect setting

andrecuellar:
## Context

PR #281 added a "kill app on disconnect" feature. I left a comment there explaining the conflicts with my auto-connect and reconnect features:

> Personally, I believe that adding termination (or kill) logic could abruptly disrupt the lifecycle of several reconnection features, since I think there are more requests to automate connections, not disconnections (conflicts explained below)
>
> This setting is a good idea, but it directly conflicts with several of my existing reconnection and auto-start features. An unsuspecting user could get stuck in an infinite loop (app kills itself, auto-connect relaunches it, disconnect triggers kill again), or lose the ability to recover from transient connection drops.
>
> It's also worth noting that many headunits don't fully turn off — they stay powered or restart the app automatically. On those devices, having kill-on-disconnect active would completely prevent any auto-connect setting from working, since the app would kill itself before the reconnect logic has a chance to run. This could affect even more headunits than the ones we know about.
>
> **Conflicting settings:**
>
> - **Auto-connect last session** — On disconnect, `scheduleReconnectIfNeeded()` delays a few seconds then attempts to reconnect via USB or WiFi. The kill fires at 500ms, so the app dies before the reconnect can even start.
> - **Auto-connect single USB device** — Same issue. USB auto-reconnect scheduling in `AapService` is killed before it executes.
> - **Auto-start self mode** — If self mode is configured to auto-start on launch and kill-on-disconnect is enabled, the app could enter a launch → connect → disconnect → kill → relaunch cycle.
> - **WiFi connection mode (Auto/Headunit Server)** — On unclean WiFi disconnects, the app waits 2 seconds then restarts WiFi discovery. Killed at 500ms before discovery restarts.
> - **Wireless server mode** — When the head unit runs as a wireless server, the reconnect logic restarts the discovery loop after 2 seconds. Same problem.
> - **Reconnecting overlay** — `AapProjectionActivity` has a watchdog that detects frame stalls and shows a "Connection interrupted" overlay, giving the user a chance to wait for recovery. With kill-on-disconnect, the user never sees the overlay — the app just dies on any transient drop.
>
> **Additional concerns:**
>
> - `Process.killProcess()` is a hard kill that doesn't run `onDestroy()`, so `doDisconnect()` (which sends ByeByeRequest to the phone) can be interrupted mid-cleanup, leaving the phone-side connection in a bad state.
> - The kill doesn't distinguish between a clean disconnect (user unplugged) and a transient connection drop (brief signal loss). It should only trigger on clean or user-initiated disconnects.

## What this PR does (work in progress)

Starting to address the issues above. So far:

- **Warning dialog** — when the user enables "close on disconnect" and has any conflicting settings active, a dialog pops up listing them and offers to disable all of them before enabling close-on-disconnect
- **Graceful termination** — replaced `Process.killProcess()` with `finishAndRemoveTask()` + `System.exit(0)` so activities get a chance to clean up properly and `doDisconnect()` can send the ByeByeRequest to the phone

## Still to do

- Only trigger the close when the disconnect is clean or user-initiated, not on transient connection drops
- Skip the close entirely if auto-reconnect is about to kick in (make it subordinate to `scheduleReconnectIfNeeded()`)
- Rename the setting internally from "kill" to "close" for consistency
- Add translations for the new warning dialog strings
- Test on actual hardware to make sure the timing works and the ByeByeRequest goes through before the app exits


andreknieriem:
Hey @andrecuellar I merged it by accident. I want to change the kill too. That is why I not released a new beta, but I have to worked and got no time fixing it.

andreknieriem:
And the whole translations are missing too

andrecuellar:
@andreknieriem no worries, I figured it was accidental since there were no translations and the kill approach was pretty rough. I'll take care of it here. The graceful termination is already in this PR, and I'll add the translations with AI and the rest of the safeguards before marking it ready.

andreknieriem:
> @andreknieriem no worries, I figured it was accidental since there were no translations and the kill approach was pretty rough. I'll take care of it here. The graceful termination is already in this PR, and I'll add the translations with AI and the rest of the safeguards before marking it ready.

Ok. But my AA selfmode fix needs a new release soon. So if you have time to go ;) The 2.10.0 can go live later if everyhing is ok.

andrecuellar:
@Tilak-03 I've got a debug build ready for testing with the safeguards for the close-on-disconnect feature. You can grab it here:

https://github.com/andrecuellar/headunit-revived/releases/tag/debug/kill-on-disconnect-safeguards-v1

Note: you may need to uninstall the current version before installing this debug APK, since it's signed with a different key.

Let me know how it goes!

andrecuellar:
ready for review

Tilak-03:
> @Tilak-03 I've got a debug build ready for testing with the safeguards for the close-on-disconnect feature. You can grab it here:
> 
> https://github.com/andrecuellar/headunit-revived/releases/tag/debug/kill-on-disconnect-safeguards-v1
> 
> Note: you may need to uninstall the current version before installing this debug APK, since it's signed with a different key.
> 
> Let me know how it goes!

Tested the app working great, Thanks

PR #294, opened 2026-03-25
Fix MediaTek quick boot auto-start and light mode UI issues

### LINKED ISSUE #292: Auto-start on boot fails on MediaTek head units with Quick Boot / IPO (Instant Power On)
## Problem

The "Auto-start on boot" feature does not work on MediaTek-based head units that have **Quick Boot** (also known as **Fast Boot** or **Instant Power On / IPO**) enabled. This is likely the default mode on many Chinese Android head units.

When Quick Boot is enabled, the head unit does not perform a full cold boot — it **hibernates and resumes** instead. As a result:

- `BOOT_COMPLETED` is **never broadcast** because the system didn't actually boot
- `LOCKED_BOOT_COMPLETED` is **never broadcast** for the same reason
- `USB_DEVICE_ATTACHED` is **never broadcast** because the USB device was already physically connected before the system hibernated

This means **nothing triggers HeadUnit Revived to launch**, even with all auto-start and auto-connect settings enabled.

## Observed behavior

- **Cold start (after 8+ hours off):** App does not auto-start. User must open the app manually.
- **After manually opening the app once:** Disconnecting and reconnecting the USB cable triggers `USB_DEVICE_ATTACHED`, and auto-connect works correctly from that point on.
- **Logs confirm:** The app is launched from the launcher (`com.reglink.launcher4`), not from `BootCompleteReceiver` or `UsbAttachedActivity`. There is no log entry for `"Boot auto-start: received action="`, confirming the broadcast never arrived.

## Root cause

The current `BootCompleteReceiver` listens for:
- `android.intent.action.BOOT_COMPLETED`
- `android.intent.action.LOCKED_BOOT_COMPLETED`
- `android.intent.action.QUICKBOOT_POWERON`
- `com.htc.intent.action.QUICKBOOT_POWERON`

**Missing are the MediaTek-specific quick boot intents:**
- `com.mediatek.intent.action.QUICKBOOT_POWERON`
- `com.mediatek.intent.action.BOOT_IPO`

MediaTek's IPO (Instant Power On) technology is widely used on Chinese head units. When the car is turned off, the head unit hibernates instead of shutting down completely. On the next ignition, it resumes from hibernate and broadcasts its own MediaTek-specific intents — which HeadUnit Revived currently ignores.

## Test device

- **Head unit:** CMZX65-U1 (Chinese MediaTek Android 12)
- **System version:** CMZXR62N-U1_R6210_S4.26
- **MCU:** 95.05@240929 CW32F030
- **Quick Boot:** Enabled (system does not show boot logo on startup, goes directly to home screen)

## Potentially related issues

This could be the underlying cause behind several reported issues where auto-start/auto-connect doesn't work on head units:
- #147 — USB Autostart doesn't start USB Android Auto Dongle
- #173 — USB autoconnect and auto launch feature not working
- #190 — Auto launch (not auto connect)

Many of these users are likely on MediaTek-based head units with Quick Boot enabled by default, which would silently prevent all boot-triggered auto-start mechanisms from firing.

## Proposed fix

I'll open a PR adding the missing MediaTek IPO/Quick Boot broadcast actions to `BootCompleteReceiver` and the manifest. This is a small, targeted change that could resolve auto-start failures for a large number of MediaTek head unit users.

andrecuellar:
## Summary

- Add MediaTek IPO (Instant Power On) broadcast actions (`com.mediatek.intent.action.QUICKBOOT_POWERON`, `com.mediatek.intent.action.BOOT_IPO`) to `BootCompleteReceiver` so auto-start works on MediaTek head units that use hibernate/resume instead of full cold boot. Closes #292
- Fix hardcoded white text colors that were invisible in light mode: dialog title/body text, custom margin dialog inputs, and auto-connect reorder icons (drag handle, up/down arrows)
- Fix exit button text always appearing dark in light mode by excluding it from dynamic theme color updates
- Show both App theme and Android Auto night mode state in the dark mode settings subtitle
- Enable `android:supportsRtl="true"` in manifest and fix RTL layout issues in device list
- Complete missing Arabic translations and update stale ones
- Improve auto-start OEM warning to reference "this device" and mention battery/background restrictions


andrecuellar:
Ready for merge

mrkontrast-coder:
> Ready for merge

Do you have compiled apk with this PR ?

andrecuellar:
> > Ready for merge
> 
> Do you have compiled apk with this PR ?

@mrkontrast-coder compiling...

mrkontrast-coder:
I can test the MediaTek boot intent, my headunit has similar product name and MCU to the one mentioned in your post

andrecuellar:
# OUTDATED (read below)

@mrkontrast-coder here you go!

Direct download: https://github.com/andrecuellar/headunit-revived/releases/download/pr294-mediatek-lightmode-v2/com.andrerinas.headunitrevived_2.1.0.apk

If that doesn't work, grab it from the release page: https://github.com/andrecuellar/headunit-revived/releases/tag/pr294-mediatek-lightmode-v2

You'll probably need to uninstall the version you have first since this is signed with a different key.

andrecuellar:
Marking as draft — I'm adding one more thing before this is ready.

On head units that hibernate instead of shutting down, `USB_DEVICE_ATTACHED` doesn't always fire after wake, so USB auto-start doesn't work even if you plug in the phone after the radio turns on. Adding a fallback where the service starts in the background on ACC/wake intents when USB auto-start is enabled, so it can detect USB connections even when the system doesn't send the attach intent. Will push shortly.

andrecuellar:
@mrkontrast-coder hey, try this newer build instead of the previous one — I added a fix for USB auto-start on hibernating head units. Now when the radio wakes up and you have USB auto-start enabled, the service starts in the background and checks for connected USB devices, even if the system doesn't send the USB attach intent properly.

Direct download: https://github.com/andrecuellar/headunit-revived/releases/download/pr294-mediatek-lightmode-v3/com.andrerinas.headunitrevived_2.1.0.apk

Release page: https://github.com/andrecuellar/headunit-revived/releases/tag/pr294-mediatek-lightmode-v3

You'll probably need to uninstall the previous version first.

andrecuellar:
@mrkontrast-coder update — I tested on my CMZX65-U1 with Quick Boot and found the root cause. Analyzing the logs, the device's uptime was only ~2 minutes when I manually opened the app — meaning Quick Boot does a **real reboot**, not a hibernate from RAM. The process was dead after reboot, so the dynamic `SCREEN_ON` receiver was never registered and couldn't fire.

The fix: `BootCompleteReceiver` (manifest-declared, survives reboot) now also checks if "Start on screen on" is enabled and starts the foreground service after boot, which registers the dynamic `SCREEN_ON` receiver. The setting is now synced to device-protected storage so it can be read during locked boot.

## What's new in this build (v7)

### 1. Boot receiver support for "Start on screen on"
After a reboot (including Quick Boot), `BootCompleteReceiver` now starts the service when "Start on screen on" is enabled. This registers the dynamic `SCREEN_ON` receiver so it's ready for the next screen-on event.

### 2. Service auto-start on app open
When the app opens with "Start on screen on" enabled, it now immediately starts the foreground service to ensure the `SCREEN_ON` receiver is registered. Previously the service only started via USB auto-connect or the exit button flow.

### 3. New setting: "Start on screen on" (from v5)
A new toggle in Auto-start settings. When enabled, the app opens **every time the screen turns on**. On a head unit, screen on = car turned on.

Key behavior:
- If Android Auto projection was still active when the screen turned off, it **returns directly to the projection screen**
- If the connection dropped, it launches the main screen which triggers auto-connect + USB check
- The service stays alive in the background after exit to survive hibernate

### 4. Close-on-disconnect warning (from v6)
The close-on-disconnect warning dialog now includes "Start on screen on" as a conflicting setting.

### 5. Hibernate wake detection (from v5)
For users who prefer "Start on boot", the service detects hibernate wakes by tracking `SCREEN_ON` timing — if the screen was off for more than 60 seconds, it treats it as a hibernate wake.

### 6. Diagnostic logging
All wake-related system events are logged with a `WakeDetect:` prefix.

## Debug APK

**Direct download:** https://github.com/andrecuellar/headunit-revived/releases/download/pr294-screen-on-v7/com.andrerinas.headunitrevived_2.1.0_debug.apk

**Release page:** https://github.com/andrecuellar/headunit-revived/releases/tag/pr294-screen-on-v7

You'll need to uninstall the previous version first (different signing key).

## How to test

1. Install the APK
2. Go to Settings → Auto-start settings
3. Enable **"Start on screen on"** (and optionally "Auto-start on USB")
4. Open the app so the service starts (you'll see the notification)
5. Turn off the car → wait at least 2 minutes → turn it back on
6. The app should open automatically when the screen comes on
7. **Export logs regardless** (Settings → Debug → Export Logs) — the `WakeDetect:` entries will tell us what happened

The key difference from v5/v6: after the car reboots, `BootCompleteReceiver` now starts the service so the `SCREEN_ON` receiver is active when the screen turns on.

mrkontrast-coder:
> @mrkontrast-coder update — I tested on my CMZX65-U1 with Quick Boot and found the root cause. Analyzing the logs, the device's uptime was only ~2 minutes when I manually opened the app — meaning Quick Boot does a **real reboot**, not a hibernate from RAM. The process was dead after reboot, so the dynamic `SCREEN_ON` receiver was never registered and couldn't fire.
> 
> The fix: `BootCompleteReceiver` (manifest-declared, survives reboot) now also checks if "Start on screen on" is enabled and starts the foreground service after boot, which registers the dynamic `SCREEN_ON` receiver. The setting is now synced to device-protected storage so it can be read during locked boot.
> 
> ## What's new in this build (v7)
> 
> ### 1. Boot receiver support for "Start on screen on"
> After a reboot (including Quick Boot), `BootCompleteReceiver` now starts the service when "Start on screen on" is enabled. This registers the dynamic `SCREEN_ON` receiver so it's ready for the next screen-on event.
> 
> ### 2. Service auto-start on app open
> When the app opens with "Start on screen on" enabled, it now immediately starts the foreground service to ensure the `SCREEN_ON` receiver is registered. Previously the service only started via USB auto-connect or the exit button flow.
> 
> ### 3. New setting: "Start on screen on" (from v5)
> A new toggle in Auto-start settings. When enabled, the app opens **every time the screen turns on**. On a head unit, screen on = car turned on.
> 
> Key behavior:
> - If Android Auto projection was still active when the screen turned off, it **returns directly to the projection screen**
> - If the connection dropped, it launches the main screen which triggers auto-connect + USB check
> - The service stays alive in the background after exit to survive hibernate
> 
> ### 4. Close-on-disconnect warning (from v6)
> The close-on-disconnect warning dialog now includes "Start on screen on" as a conflicting setting.
> 
> ### 5. Hibernate wake detection (from v5)
> For users who prefer "Start on boot", the service detects hibernate wakes by tracking `SCREEN_ON` timing — if the screen was off for more than 60 seconds, it treats it as a hibernate wake.
> 
> ### 6. Diagnostic logging
> All wake-related system events are logged with a `WakeDetect:` prefix.
> 
> ## Debug APK
> 
> **Direct download:** https://github.com/andrecuellar/headunit-revived/releases/download/pr294-screen-on-v7/com.andrerinas.headunitrevived_2.1.0_debug.apk
> 
> **Release page:** https://github.com/andrecuellar/headunit-revived/releases/tag/pr294-screen-on-v7
> 
> You'll need to uninstall the previous version first (different signing key).
> 
> ## How to test
> 
> 1. Install the APK
> 2. Go to Settings → Auto-start settings
> 3. Enable **"Start on screen on"** (and optionally "Auto-start on USB")
> 4. Open the app so the service starts (you'll see the notification)
> 5. Turn off the car → wait at least 2 minutes → turn it back on
> 6. The app should open automatically when the screen comes on
> 7. **Export logs regardless** (Settings → Debug → Export Logs) — the `WakeDetect:` entries will tell us what happened
> 
> The key difference from v5/v6: after the car reboots, `BootCompleteReceiver` now starts the service so the `SCREEN_ON` receiver is active when the screen turns on.
> 

Works great for me, app lunches after headunit quick boot 😃. Will provide logs later

andrecuellar:
> Works great for me, app lunches after headunit quick boot 😃. Will provide logs later

@mrkontrast-coder It doesn't work yet on my headunit if quick boot is enabled :(

 I'm going to install tracking apps until I understand what's going on; I have a feeling that Quickboot is omitting to send all kinds of signals.

mrkontrast-coder:
> > Works great for me, app lunches after headunit quick boot 😃. Will provide logs later
> 
> @mrkontrast-coder It doesn't work yet on my headunit if quick boot is enabled :(
> 
>  I'm going to install tracking apps until I understand what's going on; I have a feeling that Quickboot is omitting to send all kinds of signals.
> 





Try to disable this, it might be killing stuff in the sleep state.

andrecuellar:
> Try to disable this, it might be killing stuff in the sleep state.

My headunit doesn't have that. I'll review other options in the next 24 hours to see what other events I can capture or improve, and if I don't find anything else, I'll release it as is.

andreknieriem:
I've merged together this pr with my hotspot pr and latest pr from others. 
https://github.com/andreknieriem/headunit-revived/releases/tag/v.2.1.1-beta1

mrkontrast-coder:
> > Try to disable this, it might be killing stuff in the sleep state.
> 
> My headunit doesn't have that. I'll review other options in the next 24 hours to see what other events I can capture or improve, and if I don't find anything else, I'll release it as is.
> 

I don't know if i changed something, but the autostart on quick boot also doesn't work anymore for me, those MediaTek divaces are very weird. Looking forward to any idea what might work. 💀

andrecuellar:
> I don't know if i changed something, but the autostart on quick boot also doesn't work anymore for me, those MediaTek divaces are very weird. Looking forward to any idea what might work. 💀

@mrkontrast-coder It occurs to me that these devices decide not to send events that could delay the start-up; they must have a very short timeout. In other words, if it doesn't do it quickly, it's better not to do it at all.

I'm analyzing some logs I generated; however, if I don't find anything, I'll dedicate a few hours of my weekend to tracking the quick boot with external applications or even with ADB by connecting the headunit to my computer.

andrecuellar:
Marking this as ready for review, the light mode fixes, RTL support, Arabic translations, and the OEM boot/ACC intent coverage are all solid and worth merging on their own.

However, this PR does **not** fully solve the Quick Boot auto-start problem on all head units. On some MediaTek devices (confirmed on my CMZX65-U1), none of the registered boot/ACC intents are delivered after a Quick Boot. not even `BOOT_COMPLETED`. The "Start on screen on" feature works correctly when Quick Boot is disabled (full boot with boot animation), but on Quick Boot the process is killed and no broadcast wakes it up, so the dynamic `SCREEN_ON` receiver is never registered.

I'll open a separate issue and PR specifically targeting this. I'm currently investigating what broadcasts these devices actually send during Quick Boot via ADB diagnostics.

andrecuellar:
## Quick Boot diagnostic update

Posted detailed diagnostic findings on issue #292: https://github.com/andreknieriem/headunit-revived/issues/292#issuecomment-4145839360

**TL;DR:** The Quick Boot on the CMZX65-U1 is actually a full reboot that happens with the screen off (while ACC is off). HUR _does_ start successfully via `BOOT_COMPLETED`, but ~60-90 seconds later, the Reglink firmware enters a kind of "battery saving mode" — it sees no user activity (ACC off, screen off) and force-stops all third-party apps to save power while running on residual battery. This sets the package to `stopped=true`, which blocks all implicit broadcasts — so when the user finally turns the car on, nothing can wake HUR up.

This confirms that the Quick Boot auto-start problem is a firmware-level power management issue, not a missing broadcast problem, and needs a separate dedicated investigation. The rest of this PR (light mode fixes, RTL, Arabic translations, OEM boot intents) remains solid and independent of this issue.

andrecuellar:
## Test APK v9 — Wake lock resilience + built-in Quick Boot diagnostics

Based on the diagnostic findings ([see analysis](https://github.com/andreknieriem/headunit-revived/issues/292#issuecomment-4145839360)), I've added both a defense mechanism and a built-in diagnostic logger.

### Defense: Wake lock resilience
- **PARTIAL_WAKE_LOCK** (10min timeout): Acquired when the service starts from boot, screen-on, or USB check. Keeps the CPU active and signals to the MediaTek power management that the service needs to stay alive.
- **`stopWithTask="false"`**: Prevents the service from being killed when the task is removed from recents.
- **`onTaskRemoved()` handler**: Attempts to restart the service if the task is removed.

### Diagnostics: Built-in heartbeat logger
When the service starts from boot or screen-on, it writes a heartbeat file to **`/sdcard/headunit-quickboot-diag.txt`** every 2 seconds. Each line records timestamp, uptime, ACC signal status, and wake lock state.

The file survives even after the firmware kills the app — just open it from the radio's file manager to see exactly when HUR was killed. **No ADB or PC connection needed.**

Example output:
```
=== HUR Quick Boot Diagnostic ===
Boot start: 18:28:50.123
Uptime: 30.5s
Boot reason: power_key
ACC signal: 0
WakeLock held: true
Battery opt exempt: false
PID: 2123

=== Heartbeat (every 2s) ===
Time         | Uptime   | ACC | WakeLock | Notes
-------------|----------|-----|----------|------
18:28:50.125 | 30.52    | 0   | yes      | tick=0
18:28:52.126 | 32.52    | 0   | yes      | tick=1
...
18:30:14.130 | 114.52   | 0   | yes      | tick=42
                            ← last line = when HUR was killed
```

### How to test
1. Install the APK
2. Enable "Start on boot" or "Start on screen on"
3. Turn off the car, wait >2 minutes
4. Turn on the car
5. Open file manager → Internal Storage → `headunit-quickboot-diag.txt`
6. Share the file (or take a screenshot) — the last heartbeat line shows exactly when HUR was killed

### Download
https://github.com/andrecuellar/headunit-revived/releases/tag/test/pr294-wakelock-v9

andrecuellar:
## Update: Quick Boot diagnostic results + warning banner

### Diagnostic conclusion

The built-in heartbeat logger confirmed that **the kill happens during the shutdown sequence, not after the reboot**. The Reglink firmware force-stops all third-party apps ~20 seconds after ACC goes off, right before rebooting. Wake locks and foreground service protections are ignored. Full details in [issue #292](https://github.com/andreknieriem/headunit-revived/issues/292#issuecomment-4146909537).

**For most users, the solution is to disable Quick Boot** in the head unit's system settings. For advanced users, installing HUR as a system app (root required) would bypass the force-stop.

### Changes in latest push

- **Reverted** the built-in diagnostic heartbeat logger (was only needed for testing)
- **Added Quick Boot warning banner** to the auto-start settings screen, informing users that Quick Boot may not be compatible with auto-start features
- Translations added for all 13 locales
- The wake lock resilience commit (`8e6d6ce`) is kept since it's still useful for softer kill scenarios on other devices

I'll continue investigating locally, but this PR is ready for review as-is.

andrecuellar:
Quick update on the Quick Boot investigation — I compared our app with the 360 camera app (com.baony.avm360) that comes pre-installed on my head unit and always auto-starts fine after Quick Boot.

The difference is simple: the camera app is a system app (`/system/app/Reglink_Avm360`, uid 1000, SYSTEM flag). After Quick Boot it has `stopped=false`. HUR is a user app (`/data/app/`, uid 10106) and gets `stopped=true`.

The firmware only force-stops user-installed apps during shutdown. System apps are exempt. So for users with root, installing HUR to `/system/priv-app/` would fully solve the Quick Boot auto-start issue. I'm testing this on my unit now.

For everyone else, the recommendation stays the same: disable Quick Boot if auto-start doesn't work. The warning banner in this PR covers that.

andrecuellar:
Found the root cause — it's **DuraSpeed** (MediaTek's background process manager). Adding HUR to the DuraSpeed whitelist fixes auto-start after Quick Boot completely. On my head unit the DuraSpeed settings were hidden and I needed an app called "Hidden Settings" to access them.

Updated the auto-start warning banner to mention DuraSpeed and the Hidden Settings workaround. Translations updated for all locales.

This is a much simpler solution than disabling Quick Boot or rooting the device, and it should work on most MediaTek-based head units.

andrecuellar:
Quick Boot investigation is done. The auto-start warning banner now covers everything users need to know — DuraSpeed whitelist, Quick Boot toggle, and the Hidden Settings workaround.

If none of those work, the only remaining option is installing HUR as a system app, which requires root access and some technical knowledge. At this point the ball is in each user's court since the force-stop comes from the firmware itself and there's nothing we can do from the app side.

This PR is ready for review with the current changes.

andrecuellar:
Follow-up: even installing HUR as a system app in `/system/priv-app/` doesn't survive Quick Boot on Reglink firmware. The firmware has an internal "skipped list" whitelist — only apps in that list or running as uid 1000 (platform-signed) are spared from force-stop during boot. Since we can't sign with the platform key, this is completely out of our hands.

For non-technical users, the only reliable solution remains **disabling Quick Boot** in the head unit's system settings. The DuraSpeed whitelist helps in some cases but is inconsistent. The warning banner in this PR covers all the options.

andrecuellar:
Found the Reglink firmware's private whitelists — three text files in \`/system/etc/\` that control which apps survive Quick Boot. Full details in the issue: https://github.com/andreknieriem/headunit-revived/issues/292#issuecomment-4150981513

Testing adding HUR to all three lists now. If it works I'll write a guide for users with root access on Reglink head units.

andrecuellar:
IT FINALLY WORKS!! 🎉🚀

After countless hours of investigation — diagnostic dumps, heartbeat loggers, system app conversion, DuraSpeed discovery, and finally finding the Reglink firmware's hidden whitelists — auto-start on Quick Boot is working perfectly on my CMZX65-U1!

The fix was adding HUR to three whitelist files in `/system/etc/`. Full details and commands in the issue: https://github.com/andreknieriem/headunit-revived/issues/292#issuecomment-4151016109

Can't fix this from code unfortunately — it's a per-device firmware thing. But at least now we know exactly what's going on and how to solve it for users with root access. For everyone else, disabling Quick Boot remains the simplest option.

What a ride 🚀

andreknieriem:
Wow! What a great success! Will you make a small tutorial? I can add it to the wiki and the webseite. Both are here on github and are working with md-files. :)

andrecuellar:
> Wow! What a great success! Will you make a small tutorial? I can add it to the wiki and the webseite. Both are here on github and are working with md-files. :)

I will do a tutorial this weekend :)

PR #300, opened 2026-03-26
Translation Correction (Brazilian Portuguese)

tsabaia:
Translation Correction (Brazilian Portuguese)


andreknieriem:
Thank you. All new labels are translated via AI. Better with native speakers ;)

PR #364, opened 2026-04-10
Fix: Improve focus management and physical button handling in Self Mode

Korvix:
## Context
I’ve implemented a fix for issues encountered when using the **Self Mode** configuration with physical (Bluetooth) controllers. I decided to base my changes on the `v2.1.1` tag, as it was the most recent version that remained stable and functional for my setup.

## Changes
* **KeyEvent Handling**: Replaced standard `onKeyDown` and `onKeyUp` handling with `dispatchKeyEvent`. Previously, only the side arrows were working, and the Enter/D-Pad Center button was not being detected because it wasn't captured on standard key events. This change ensures that all button presses are correctly intercepted.
* **Focus Management**: Modified the focus logic to ensure it stays locked on the top-level application (Android Auto projection). 
    * This allows for immediate use of physical buttons right after the app starts.
    * It prevents focus loss that occasionally occurred when navigating through or exiting menus.

These changes significantly improve the stability and usability of the interface when using external Bluetooth controllers.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapProjectionActivity.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapProjectionActivity.kt
@@ -167,9 +167,14 @@ class AapProjectionActivity : SurfaceActivity(), IProjectionView.Callbacks, Vide
         }
 
         setContentView(R.layout.activity_headunit)
+        takeKeyEvents(true)
+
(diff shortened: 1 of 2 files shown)


andreknieriem:
Hey @Korvix . First of all. Thank you very much for your PR :) I will take a look at these changes

andreknieriem:
Sorry, that I've never merged this. I don't thing the focus needs to be so aggressive. I will close this for now. Feel free to contribute something in the future :)

PR #365, opened 2026-04-11
Create configurable audio queue and audio buffer in settings

irwanrhmn:
I've noticed a few spots in AudioTrackWrapper.kt where hardcoded values were causing some audio lag for some users. This PR makes those values configurable so people can tune the performance to their specific setup.

1. Audio Queue Limit (Line 33)
Right now, the audio queue is essentially bottomless. For a lot of users, this leads to a massive delay where the audio falls way behind the action. I’ve made the queue limit configurable. I kept the default behavior exactly as it is now so nothing changes for the average user, but power users can now tighten this up in the Audio settings to cut down that latency.

2. Buffer Size (Line 281)
We’ve been using a hardcoded buffer size of 8. It’s great for stability, but it creates a noticeable gap between a user’s interaction and the actual sound coming out. Like the queue limit, I’ve made this adjustable. It still defaults to 8 to keep things stable out of the box, but users can now lower it if they want a more responsive, "snappier" experience.

Overall, this should give users a lot more control over their audio sync without breaking the experience for everyone else. I test this with Audio Queue Limit to 20, and Buffer Size to 2 on my chinese headunit.

Audio Queue 20, Buffer 2x

https://github.com/user-attachments/assets/06d8d871-a6a4-4a4f-bd50-ebc04e04547d

Audio Queue Unlimited, Buffer 8x

https://github.com/user-attachments/assets/36381df0-089e-4c61-8fd8-9962383a09e5

Thank You!

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapAudio.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapAudio.kt
@@ -126,8 +126,8 @@ internal class AapAudio(
             }
             val gain = (1.0f + (offset / 100.0f)).coerceIn(0.0f, 2.0f)
 
-            AppLog.i("AudioDecoder.start: channel=$channel, stream=$stream, gain=$gain, sampleRate=${config.sampleRate}, numberOfBits=${config.numberOfBits}, numberOfChannels=${config.numberOfChannels}, isAac=${settings.useAacAudio}")
-            audioDecoder.start(channel, stream, config.sampleRate, config.numberOfBits, config.numberOfChannels, settings.useAacAudio, gain)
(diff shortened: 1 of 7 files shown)


andreknieriem:
Thank you so much! Can you look at code assists comments and maybe fix them if tjey are good point? :)

irwanrhmn:
Hi @andreknieriem 

I'll take a look tomorrow, is it ok? I'm in the middle of traffic jam right now 😅

andreknieriem:
Haha of course. I am not at home too and struggling with beta3 tomorrow.

irwanrhmn:
Hi @andreknieriem ,

Checking the gemini suggestion, it does make sense since we will get a crash instead of skipping the audio frame. I change the logic to offer, and test it with my (beloved, haha) chinese headunit. No issues while testing it.

i included it in the latest refactor commit, and here is the latest debug log from the app.

[HUR_Log_20260412_131140.txt](https://github.com/user-attachments/files/26654722/HUR_Log_20260412_131140.txt)

PR #378, opened 2026-04-13
Removing native ssl, with very old open ssl 1.1

andreknieriem:
This will reduce file size by 5mb and removes unnecessary bloat of the code.


DekkerlionTSSF:
Hi, will this affect legacy devices, is there still any performance benefit from using native ssl on older devices ? Let's say a Snapdragon 450 or 625 ? If yes and the difference is massive, would you also be able to provide a separate apk with support for these? I'm not sure, I'm still testing around with my head unit what works best, it's a "Lenovo D1 Pro" with MSM8953 chip.

andreknieriem:
This shouldn't affect the performance and is much more reliable and debuggable. I can remove hundrets of files with an very old version of openssl 1.1

PR #383, opened 2026-04-14
Added calibration overlay

andreknieriem:


andreknieriem:
I think this is not needed. I left it out, but the Code can be recovered if needed

PR #502, opened 2026-05-13
Updated handling navigation.proto 

Bastel2020:
Fixes navigation broadcast issues by switching to new message handling

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapNavigation.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapNavigation.kt
@@ -1,12 +1,8 @@
 package com.andrerinas.headunitrevived.aap
 
-import android.app.NotificationChannel
-import android.app.NotificationManager
-import android.app.PendingIntent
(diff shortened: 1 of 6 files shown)


andreknieriem:
Nice! And this works? Code Assists says some things. Can you check them?

evgley:
@andreknieriem ,  @Bastel2020 I've applied this to 2.3.0-beta2 . All conflicts resolved using files from this PR
Notifications came back, distance works.

The only thing I've found that is something wrong with intents while route is abount to finish (finish flag is shown in AA interface).  But it works fine on current 2.3.0-beta2 (after fixes from @andreknieriem and BEFORE merging this PR)

Thanks for your work !

andreknieriem:
@Bastel2020 Yeah first thing from code assists say destination might not change. Maybe it is this wrong proto?

Bastel2020:
Yes, that could be the case. I used a leaked proto description from a Polish website, and it might contain errors. Unfortunately, I can't verify that destination works yet.

PR #549, opened 2026-05-27
feat: implement back key routing and add keymap for back key

JanRi3D:


andreknieriem:
Can you look at the points code review says? Thank you :)

PR #574, opened 2026-06-03
Complete Spanish translations and align Self Mode wording

andrecuellar:
Spanish locales were missing 41 strings, filled them in.


andrecuellar:
Heads up: I did the Spanish translations in this PR by hand since I am a native speaker. The other locales are coming in a separate AI-assisted PR.

PR #602, opened 2026-06-14
fix: USB device list duplicates and Android Auto projection launch on Android 10+

jeancarloscc:
## Problem

  Three USB connectivity bugs affecting head units:

  1. **Android Auto not appearing on screen** — `launchAapProjectionActivity()` called
     `startActivity()` directly from a background Service. On Android 10+, this is
     silently blocked by the OS. The phone showed "connected" but the head unit never
     opened the projection screen.

  2. **Duplicate entries in USB device list** — During AOA mode switching, the phone
     Also, some head units have internal components (Bluetooth module, GPS, etc.) that
     identify with the same USB strings as the connected phone, creating false duplicates.
     
  3. **`onNewIntent()` using stale intent** — `resolveUsbDevice(getIntent())` was
     called instead of `resolveUsbDevice(intent)`, causing the wrong USB device to be
     resolved when the phone re-enumerated in accessory mode.
     
  ## Fix
  
  - `AapService`: apply overlay trampoline in `launchAapProjectionActivity()`, same
    technique already used in `launchMainActivityOnBoot()` (new `ActivityLaunchPolicy`
    extracted for testability)
  - `MainViewModel`: deduplicate device list by `uniqueName` and hide accessory-mode
    entries when a normal-mode entry for the same device is also present
  - `UsbDeviceCompat`: include VID:PID in `uniqueName` so devices with identical
    USB strings but different hardware are treated as distinct
  - `UsbAttachedActivity`: fix `onNewIntent()` to use the `intent` parameter
  - `UsbListFragment`: remove kernel device path from display (changes on every
    reconnect, confuses users)
  - `VideoDecoder`: add missing `MediaCodecInfo` import
  
  ## Tests
  
  Unit tests added for all extracted pure logic following existing `ProjectionKeyPolicy`
  pattern (JUnit4, no mocking framework). 
  
  ## Tested on
  
  H-TECH car multimedia unit (MediaTek, Android 10)

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
@@ -935,10 +935,51 @@ class AapService : Service(), UsbReceiver.Listener {
             AppLog.i("AapService: Skipping projection launch because PiP is active")
             return
         }
-        startActivity(AapProjectionActivity.intent(this).apply {
+
+        val intent = AapProjectionActivity.intent(this).apply {
             putExtra(AapProjectionActivity.EXTRA_FOCUS, true)
             addFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT)
-        })
+        }
+
+        val canOverlay = Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && AndroidSettings.canDrawOverlays(this)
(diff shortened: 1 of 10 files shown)


andreknieriem:
Hey nice PR, can you please look at the things copilot and gemini suggests?

jeancarloscc:
Thanks for the feedback! I've made the requested adjustments and pushed them in the latest commit.

PR #638, opened 2026-07-04
Added support for manual BSSID configuration in wireless settings for Native mode.

rakshan-kumr:
Due to strict Android privacy restrictions regarding local network identifiers, the Headunit app cannot programmatically retrieve the Wi-Fi Direct BSSID while running in Native mode. This blocks the seamless automated connection flow.

To bypass this limitation and restore connectivity, this PR introduces a fallback manual configuration feature. Users can now manually retrieve the BSSID (e.g., via ADB or network tool applications) and input it directly into the Headunit app's settings. The app will then use this user-provided BSSID to establish a successful connection with the Android device.


andreknieriem:
This is really great idea :)

andreknieriem:
It seems you forgot: DialogUtils. This class is not included.

Issue #650, opened 2026-07-10
Black screen on startup by wi-fi ssid connection


ViBorges:
I put the ssid for my phone hotspot on autostart, when I get the notification I click on the wi-fi icon and it connects but the screen most of the time is black. If I press home on my headunit and go back to HURev the android auto screen is back up again. Also when it connects my phone is locked. Sometimes it works fine but most of the time it starts with a black screen. Using 3.1.1-beta.

andreknieriem:
Ok. Can you export some logs and post it here?

jeffdapaz:
Hi @ViBorges and @andreknieriem,  

I’d like to add a few notes. I’ve been experiencing the black screen issue at startup as well, but with some slight differences.  

I connect via USB, and when Android Auto starts, after two or three minutes the screen goes black while the audio keeps working. However, as @ViBorges mentioned, if I go to the home screen or simply put HUR in the background and then return to it, Android Auto displays the screen again.  After that I don't have more black screens, only if I restart my head unit.

ViBorges:
I was trying to reproduce it on this weekend to get some logs but this time it was working, maybe mostly of the time that happens was because the screen was switching to reverse camera before I connect to HUR. Tried to reprocude on 3.1.0 but it was working fine.  When I get some time I'll get the logs.

jeffdapaz:
Hi @andreknieriem,

I managed to capture a log while the black screen was happening. I'm attaching the log file to this issue. When the black screen occurred, I performed the workaround of going to Home and back to the app to restore the screen, so both the failure and the recovery are visible in the log.

I asked an AI assistant to analyze it, and here is what was found.

## The black screen phase

After the first frame is rendered ("Hiding loading overlay after first video frame" at 10:00:38.483), the log immediately enters a flood of dropped frames on the video surface (SurfaceTexture-1-1054-0):

queueBuffer: slot 0 is dropped
queueBuffer: slot 2 is dropped
queueBuffer: slot 3 is dropped
... (hundreds of dropped frames)

And critically, in the middle of it:

Long monitor contention with owner GLThread 78 ... nativeUpdateTexImage() ... for 191ms
avc: denied { ioctl } for comm=GLThread path="/proc/mtk_mdp_cmdq"

[...]

PR #673, opened 2026-07-15
Recover automatically from post-first-frame video display stalls (#650)

andrecuellar:
## Problem

On some head units the projection freezes shortly after the first video frame, often on the Android Auto boot logo, while audio keeps playing. Pressing Home and reopening the app (which rebuilds the surface) restores it. Reported in issue #650.

## Root cause

Two layers, confirmed from the logs and the code history:

1. Sampling the hardware decoder's external SurfaceTexture in-app via GLES (`updateTexImage` on the GL thread) forces a per-frame color conversion on some SoCs. On MediaTek this goes through the MDP (`mtk_mdp_cmdq`, with SELinux denials) and is very slow, up to ~2s per frame, so playback collapses to 2-5fps and looks frozen. A fresh surface clears it, which is why Home and reopen works. TextureView and SurfaceView let the system composite the decoder output directly and avoid that path.

2. Affected devices end up on GLES because the auto-optimize wizard recommended GLES for every Android 5.0-8.1 device (`SystemOptimizer`), regardless of SoC or hardware-decode capability. This is not a recent regression, the recommendation predates 3.1.0.

## Change

Fix at the source and add a safety net for installs already on GLES:

- `SystemOptimizer` now recommends a direct SurfaceView for 5.0-8.1 devices that can decode video in hardware, keeping GLES only for devices without hardware HEVC (which may need the software YUV sink GLES provides).
- Runtime recovery for existing GLES installs: a watchdog detects the stall (the phone is still streaming but the display stopped drawing or collapsed to a few fps, using a per-view long-frame counter), rebuilds the surface, and if a device keeps stalling it switches that session to SurfaceView. It is skipped while the bundled software HEVC decoder is active, and never fights a phone-side pause.
- Diagnostic logging of the recommended and active view mode alongside the SoC, so the path can be confirmed from device logs.

## Testing

Builds clean on the github debug flavor. One tester confirmed the SurfaceView path fixes their case; iterating with another MediaTek reporter via debug builds shared in issue #650.


MrEAlderson:
Potential relation with https://github.com/andreknieriem/headunit-revived/pull/646. Changes from that #646 may be reverted if this PR applies a fix universally well already, which I could test

andreknieriem:
So this is only a PR for fixing it with auto surface destroy and you are trying to find, why this is even happening, right?

andrecuellar:
Yeah, but I did find the cause: on MediaTek the GLES path runs the decoder through the MTK MDP (`mtk_mdp_cmdq`), which collapses to 2-5fps and looks frozen. So it now detects that and, if it keeps happening, switches the session to SurfaceView to skip that path. Still draft until it's confirmed on a real unit.

DekkerlionTSSF:
Hi does this affect MTK devices only or can a Snapdragon 450 (MSM8953) also benefit from it ?

andrecuellar:
@DekkerlionTSSF not MTK only. It reacts to the symptom (video stalls while the phone keeps streaming), not to the chip, so any device where the GLES path freezes like that should benefit, a Snapdragon 450 included if it shows the same thing. MediaTek is just where I nailed down the exact cause. Quick check on any device: if it freezes, set View mode > SurfaceView in Settings, and if that fixes it, this just automates it.

PR #677, opened 2026-07-20
Transport.on quit

notathf:
Changed onQuit handling to keep the active transport reference available while executing the disconnect sequence. _transport is now cleared only after transportedQuited() finishes, preventing race conditions during audio/video decoder shutdown.


andreknieriem:
I've fixed that already in version 3.2.0-beta1 ;)

Issue #647, opened 2026-07-09
Enhancment Auto Dark Mode (Time/gps) using user defined coordinates.


zd3sf:
Is there a way to enter a fixed location for sunrise/sunset for devices with no GPS? Manual time needs constant adjustment between summer and winter. If the current mechanism uses GPS location to lookup sunrise/sunset, does it need internet connection or are the sunrise/sunset values stored already on the device?

Thanks

andreknieriem:
This could be a good addition ;)

andrecuellar:
@zd3sf  I've been working on this for a few days, there will be news soon.

andrecuellar:
@zd3sf 

I put together a build that adds this.

You can now set a fixed sunrise/sunset location, either by typing the coordinates or by picking a point on a map, so head units without GPS still get correct day and night. There is also a new Location theme mode that switches light or dark by area (saved places), for both the app and Android Auto.

Test build:
https://github.com/andrecuellar/headunit-revived/releases/download/test-647-location/com.andrerinas.headunitrevived_3.2.0-beta2_debug.apk

Heads up: uninstall your current version first. It is a debug build with a different signature, so it will not install over the Play Store one, and you will lose your settings.

zd3sf:
Thank you Andre for working on this. I will be testing this build this evening!

zd3sf:
Hi Andre. I made a connection during the day and it worked and I made a connection at night and it worked. I haven't tested the transition between day and night yet.

I noticed one small thing. You have to choose fixed point and hit save, then choose coordinates on the map. If you choose coordinates on the map before saving, it will revert to automatic (GPS)

I also don't fully understand the use case for the location radius feature? Will it be auto day/night inside the circle and just one mode outside? But why? Just curious.

Thank you so much for working on this feature!

andrecuellar:
@zd3sf thanks for the thorough testing.

I fixed the revert to GPS. The problem was that the fixed point only applied when you saved the settings screen, but the map saved the coordinates separately. Now saving the point on the map (or typing the coordinates) turns on the fixed point directly, so it no longer falls back to GPS.

About the radius: you were right, it did not belong on the sunrise/sunset point. That point is just a single coordinate used to compute sunrise and sunset, so I removed the radius and the circle there. The radius only exists for the saved places in the new Location theme mode, which do depend on your device location: inside a place you get the light or dark you set, outside you get your chosen default. That is a separate feature from the fixed point.

[...]

PR #710, opened 2026-07-27
Manual coordinate night mode and geofenced day/night

andrecuellar:
Adds a Manual Coordinates option to the Android Auto night mode so head units without GPS can still get correct sunrise/sunset by picking a fixed point on a map (issue #647).

It also extends this into saved locations like Home or Work. Each location is a point plus a radius and can force day or night while you are inside it, and can optionally limit auto-start and auto-connect to that area. Position comes from the head unit GPS or the GPS the phone forwards during projection. If no location is available the automation is allowed, so devices without GPS are never blocked.

The map is OpenStreetMap through Leaflet in a WebView. It only needs internet while picking a location, and there is a notice about that before it opens.

Opening as draft while I test it on device.


andrecuellar:
Marked ready for review.

If you want to try it, here is a debug build:
https://github.com/andrecuellar/headunit-revived/releases/download/test-647-location/com.andrerinas.headunitrevived_3.2.0-beta2_debug.apk

Heads up: uninstall your current version first. This is a debug build with a different signature, so it will not install over the Play Store one, and you will lose your settings.

PR #714, opened 2026-07-28
Fixes chip busy at HURev startup

notathf:
https://github.com/andreknieriem/headunit-revived/pull/695/commits/ebab63a8eabfe08cf8489459b10ba9dc0bd094fa

fix issues from previous unsuccessful CI runPrevious PR checks were cancelled or failed. This commit addresses those issues to allow merging.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/connection/WifiDirectManager.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/connection/WifiDirectManager.kt
@@ -511,7 +511,7 @@ class WifiDirectManager(private val context: Context) : WifiP2pManager.Connectio
             })
         } catch (e: Exception) {}
 
-        // 1. Stop any ongoing discovery and remove group to start fresh
+        // 1. Stop any ongoing discovery
         mgr.stopPeerDiscovery(ch, object : WifiP2pManager.ActionListener {
             override fun onSuccess() { checkGroupAndCreate() }
             override fun onFailure(reason: Int) { checkGroupAndCreate() }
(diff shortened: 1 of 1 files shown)


andrecuellar:
Wandered in out of curiosity to see what this PR does. Heads up: besides dropping removeGroup, the diff also deletes the `lastKnownBssid = null` line right next to it. That guard was added just a couple of days ago (the "Clear stale lastKnownBssid on every removeGroup() call site" change) so a stale BSSID from the previous group is not handed to the phone, since createGroup() assigns a new random MAC each time.

The description only mentions removeGroup / CI, so dropping that guard looks accidental. If the goal is just to remove removeGroup for the chip-busy issue, it would be safer to keep `lastKnownBssid = null` in the branch where a new group gets created, so this does not re-open the stale-BSSID problem.

notathf:
removeGroup is not called, used established wifi-direct connection. Is it necessary to reset lastKnownBssid if the group is created only when the app is launched?

andreknieriem:
@notathf, so removing it shpuld not cause other issues?

notathf (commented):
Do not clear `lastKnownBssid` when checking for an existing Wi-Fi Direct group. 
Reset `lastKnownBssid` it only before creating a new group, since the cached BSSID remains valid for the current group and is still needed as a fallback when Android returns a masked BSSID.

notathf:
> @notathf, so removing it shpuld not cause other issues?

added reset `lastKnownBssid` it only before creating a new group, since the cached BSSID remains valid for the current group and is still needed as a fallback when Android returns a masked BSSID.

PR #717, opened 2026-07-29
Fix projected touch coordinate normalization

vivekpokale21:
## Summary

Fix projected Android Auto touch forwarding by normalizing touch coordinates against the actual `OverlayTouchView` dimensions instead of assuming `HeadUnitScreenConfig.getUsableWidth()` / `getUsableHeight()` always match the input surface.

Found while testing a projected Android Auto session (Galaxy S25+ to Galaxy S20 FE, wireless over phone hotspot, TextureView). YouTube Music's home screen playback controls were the easiest place to see the issue: the touch overlay was laid out at `2400x1080`, while `HeadUnitScreenConfig` reported a smaller derived usable size (`2312x1035` observed during repro), producing a consistent offset. Large targets like app icons absorbed it fine; YT Music's compact play/previous/next buttons did not. After normalizing against the actual `OverlayTouchView` size, the same controls became reliable.

## Details

- Use the measured touch overlay width/height as the primary coordinate space for projected touch mapping.
- Fall back to the projection view size, then `HeadUnitScreenConfig`, if the overlay has not been measured yet.
- Move the coordinate math into `TouchCoordinateMapper` so the normalization behavior is unit-testable.
- Preserve existing negotiated margin handling: unchanged, still applied in `TouchCoordinateMapper.kt` before the final scale step.
- Preserve HUD mirroring behavior: the X-flip for mirrored HUD output happens before coordinates enter the new mapper, so mirroring logic itself is untouched.
- Keep the existing forced-scale `stretchToFill` inversion behavior: the mapper reads the same `stretchToFill` / `forcedScale` flags from `HeadUnitScreenConfig` and applies them in the same order as before, just against the corrected measured touch dimensions.

## Test plan

- [x] `./gradlew :app:testGithubDebugUnitTest`
- [x] `./gradlew :app:assembleGithubDebug`
- [x] Manual projected Android Auto touch test (Tested on S20FE, touches land reliably)
  - Start a projected Android Auto session on the tested device setup.
  - Tap YouTube Music home screen playback controls repeatedly.
  - Confirm taps land reliably without offset tapping.
  - Confirm larger projected controls still behave normally.

## Risks / compatibility

Applies to shared projected Android Auto touch forwarding, not one specific projected app. Should improve alignment wherever the actual input overlay dimensions differ from what was previously used for normalization. Devices where those already match should behave the same, aside from very small rounding differences.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapProjectionActivity.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapProjectionActivity.kt
@@ -1195,13 +1195,25 @@ class AapProjectionActivity : SurfaceActivity(), IProjectionView.Callbacks, Vide
             return
         }
 
+        val view = projectionView as View
+        val effectiveFullscreenMode = activityFullscreenOverride ?: settings.fullscreenMode
+        val measuredTouchSurfaceEnabled = settings.useMeasuredTouchSurface &&
+            effectiveFullscreenMode == Settings.FullscreenMode.IMMERSIVE
         val overlay = touchOverlayView
-        val viewW = (overlay?.width ?: 0).takeIf { it > 0 }?.toFloat()
-            ?: (projectionView as View).width.takeIf { it > 0 }?.toFloat()
-            ?: HeadUnitScreenConfig.getUsableWidth().toFloat()
-        val viewH = (overlay?.height ?: 0).takeIf { it > 0 }?.toFloat()
-            ?: (projectionView as View).height.takeIf { it > 0 }?.toFloat()
-            ?: HeadUnitScreenConfig.getUsableHeight().toFloat()
+        val viewW = if (measuredTouchSurfaceEnabled) {
+            (overlay?.width ?: 0).takeIf { it > 0 }?.toFloat()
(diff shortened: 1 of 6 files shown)


andreknieriem:
Hey, thanks for your PR :)  I really hate to make changes on this topic :D If you change on line it works on 2 devices more and 100 devices less. This is a really problematic topic.

I guess we should somehow make a test build or I merge it for beta3 and will take feedback. Touch and Viewsize had errors for so long and I thought everythings works now finally.

vivekpokale21:
I understand the concern. I did some digging into the `2312x1035` value on my S20 FE. It looks like `HeadUnitScreenConfig` captures `displayMetrics` during projection startup while the phone’s navbar/gesture area is still reflected in the resource metrics. Even after immersive fullscreen visually hides the bars, `resources.displayMetrics` continues to report `2312x1035`, while the actual projection/touch overlay is `2400x1080`. That mismatch is what causes the touch drift.

To reduce risk, I’ll update the PR so the new touch normalization is behind a default-off experimental checkbox. With the setting disabled, the existing touch mapping path should be unchanged. Will also narrow down the workaround to plain immersive fullscreen only.

Issue #715, opened 2026-07-28
Proposal: a smarter Settings screen and an intelligent first-run wizard


andrecuellar:
As the app grows, the Settings screen has become long. It now has around 13 sections and more than 60 options. As a daily user I realize I never touch about 70 percent of them because the defaults already work well. On top of that I connect by cable, so the whole Wireless Connection group is noise I always scroll past.

I would like to rework the Settings experience in a future version with these changes:

1. Replace the current first-run helper with an intelligent, step-by-step wizard that scans the device (panel resolution, DPI, HEVC support, light sensor) and proposes the best video, display and night-mode settings, with short help text on each step.
2. Ask up front how the user connects (USB cable direct, USB wireless adapter, or WiFi) and tailor the rest of setup and the settings visibility to that choice.
3. Show this wizard once to everyone, including existing installs after the update, presented as a new feature.
4. Add a search box to Settings that also matches individual sub-options.
5. Split Settings into Basic and Advanced. Basic shows a curated, everyday subset, Advanced shows everything.
6. When the user is on cable, keep the wireless options out of Basic (still one tap away in Advanced, or reachable by changing the connection mode), so cable users get a clean screen.

I plan to build this as a single branch on top of my currently open PRs. Feedback on scope and on which options belong in Basic is very welcome before I open the PR.

andreknieriem:
This sounds like a really good idea! A checkbox for Wifi in General would be helpful and hide all the wifi settings if not used for example. I've thought about for a long time, but never had the time. Maybe there a more "Quick-Settings" too if you are in the projection already. 

Additionally I wanted something like a feature toggle. So, I am not using wifi, remove the wifi button from the main screen, Self Mode too, etc. And a right hand side layout for eg. the exit and save button.

andrecuellar:
Thanks! I had actually been mulling over something like this for a while. About a month ago I put together a test on my fork to see how it felt, and right now I am tweaking and rearranging a few things, since the redesign is large and I would rather do it in phases.

At the moment I am deep in the settings screen part. The home screen side is on my radar for a later phase. I will keep you posted as I make progress.

andreknieriem:
I know, that there are quick settings. But they are just a list of settings I thought can help. Maybe there are some more useful settings which do not require a restart. Eg the surface change?

andrecuellar:
[...]

PR #718, opened 2026-07-29
Settings redesign and onboarding wizard

andrecuellar:
This delivers the settings and onboarding rework tracked in #715. It builds on the recently merged resolution cap, geofence night mode and apply-view-mode changes, which are now in main.

New features:

- Redesigned settings screen with Basic and Advanced tabs, plus a search box that matches settings, sub-settings and picker options.
- A new onboarding wizard that replaces the old setup dialog, with one step per choice, a "Do it later" option, and a hint on each step that the choice can be changed later in Settings. It also honors Extreme Dark mode.
- Connection-aware options: choose USB, WiFi, Self Mode or All in the wizard, and any option that does not apply to the chosen connection is hidden across settings, auto-start and auto-connect.
- A graphical DPI picker shared by the wizard and a dedicated settings sub-screen, with a live preview, a slider that reaches any real device density (up to 640 DPI), an editable value field, and Small, Medium, Large and Huge presets. The sub-screen saves on its own and shows the density that Automatic will use.
- A "Car brand" step in the wizard where you pick or type your car brand. It fills the display name and manufacturer fields that Android Auto shows in its connection history and welcome screen, with a live dashboard preview and a strip of common brands.
- Resolution is now recommended from the real panel size, with a warning when the choice is larger than the panel or needs high bandwidth.
- All new strings are translated into every supported locale.


andrecuellar:
This is ready to merge now :)

andrecuellar:
Folding in a fix for the fixed sunrise/sunset point reported in #647.

Saving the point on the map now really enables the fixed point, so it no longer reverts to GPS when the settings screen was not saved first (typing the coordinates does the same). I removed the radius from the coordinate picker, since a single reference point does not need an area (the radius stays for saved places, which depend on the device location). I also added a short help line explaining it is the reference point used to calculate sunrise and sunset.

PR #721, opened 2026-07-29
Fix Auto (sunrise/sunset) mode getting deselected when opening the map (#647)

### LINKED ISSUE #647: Enhancment Auto Dark Mode (Time/gps) using user defined coordinates.

andrecuellar:
Fixes #647.

While testing the sunrise reference point I noticed a bug: if I set the theme mode to Auto (sunrise/sunset) but had not saved it yet, then went to enter coordinates or open the map, the Auto mode got deselected. I had to save the mode first, and only then could I save the map point.

The cause is the pending changes pattern in DarkModeFragment. Opening the map navigates away and recreates the fragment, which re-reads the still unsaved mode from settings and reverts the selection.

This change commits the pending Auto mode and the fixed point source to settings before entering coordinates or opening the map, and it asks the service to re-send the Android Auto night mode. When returning from the map the picked point is reflected in the coordinate fields. The app theme is only refreshed when not navigating away, to avoid an activity recreate in the middle of the navigation.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/main/DarkModeFragment.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/main/DarkModeFragment.kt
@@ -369,6 +369,24 @@ class DarkModeFragment : Fragment(), SensorEventListener {
         }
     }
 
+    /**
+     * Persists the Auto/sunrise mode(s) and the fixed-point source immediately, so the mode
+     * is not deselected when the fragment is recreated on returning from the map (issue #647).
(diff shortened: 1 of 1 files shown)


andrecuellar:

PR #722, opened 2026-07-29
Fix WiFi Direct not auto-connecting when reusing an existing group (#719)

### LINKED ISSUE #719: Regression: wireless (WiFi Direct) no longer auto-connects, Wireless Helper must be opened by hand
Reported by @ViBorges in #650. That issue is about a black screen on startup and has nothing to do with this, so I am opening this separate issue as a favor to @ViBorges to keep the topics from getting mixed.

I do not use WiFi (I use a wireless USB adapter) and I do not touch the WiFi Direct code, so I cannot reproduce or test this. I am only writing it up because I traced where it likely comes from.

### Symptom (from @ViBorges)

With recent builds the wireless connection no longer connects on its own. It used to connect automatically without opening the Wireless Helper, and reconnect in seconds after a drop. Now the WiFi icon in HUR has to be tapped and the Wireless Helper opened by hand to connect, and the automatic reconnect is gone.

### When it started

It worked on a build from July 16 and broke on a later build. In that window the only changes to the WiFi Direct path were:

- PR #695 by @notathf ("Try fix wifi direct"): stop removing and recreating the P2P group, reuse the existing group instead.
- PR #699 by @o-jcardenass ("Fix WiFi Direct connection drops, stale BSSID caching..."): the BSSID handling (stale BSSID clearing and a reflection based fallback for the masked P2P MAC).

### Hypothesis (after reading the code)

In `WifiDirectManager.checkGroupAndCreate()`, when an existing P2P group is found the code now only logs `Existing P2P group found, skip createGroup` and returns. It does not run `onGroupInfoAvailable` for that existing group.

`onGroupInfoAvailable` is where the group SSID, passphrase, IP and BSSID are resolved (including the masked MAC fallbacks) and delivered to the phone via `onCredentialsReady(ssid, psk, ip, bssid)`. Before the change, creating a fresh group always led to `onGroupInfoAvailable` and that credential delivery, so the phone received the credentials and auto joined.

After the change the group is reused and no longer recreated, so `checkGroupAndCreate()` takes the skip branch, `onGroupInfoAvailable` never runs for the reused group, and `onCredentialsReady` is never called. The phone never receives the SSID, passphrase, IP or BSSID for the current group, so it cannot auto join. The user has to open the Wireless Helper by hand, and the silent auto reconnect is lost. The skip branch also never sets `isGroupOwner = true`.

### Possible fix

In the reuse branch of `checkGroupAndCreate()`, do not just return. Still call `requestGroupInfo(channel, this)` (and set `isGroupOwner = true`) so `onGroupInfoAvailable` runs on the existing group, resolves the BSSID and delivers the credentials to the phone, the same way the fresh group path does.

@notathf and @o-jcardenass, you made these changes and know this area far better than I do, so you can confirm the intent and the right fix.

@ViBorges, you are the one affected, so a connection log would help them a lot (Settings > Debug > Export Logs).

andrecuellar:
Fixes #719.

When an existing P2P group is present, WifiDirectManager.checkGroupAndCreate() took a branch that only logged and returned, so onGroupInfoAvailable never ran and the group credentials (SSID, passphrase, IP, BSSID) were never delivered to the phone through onCredentialsReady. Since the group is now reused instead of being removed and recreated (PR #695), the phone stopped receiving the credentials and could no longer auto-join, so the Wireless Helper had to be opened by hand and the quick auto-reconnect was lost.

The reuse branch now resolves and delivers the credentials for the existing group, the same way the fresh-group path does: it sets the group-owner flag and runs the device-info then group-info request so onGroupInfoAvailable runs. The BSSID resolution from PR #699 still applies, since it lives inside onGroupInfoAvailable.

I cannot reproduce or test this myself, since I use a wireless USB adapter and not WiFi. A test build and a request to confirm are on #719.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/connection/WifiDirectManager.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/connection/WifiDirectManager.kt
@@ -529,7 +529,24 @@ class WifiDirectManager(private val context: Context) : WifiP2pManager.Connectio
                 createNewGroup(0)
                 return@requestGroupInfo
             } else {
-                AppLog.i("Existing P2P group found, skip createGroup")
+                // The group already exists (reused), so no WIFI_P2P_CONNECTION_CHANGED
+                // broadcast fires to trigger onGroupInfoAvailable. Resolve the device info
+                // and request the group info ourselves so the SSID, passphrase, IP and BSSID
+                // are delivered to the phone via onCredentialsReady, exactly like the freshly
+                // created group path does. Without this the phone never receives the
+                // credentials and cannot auto-join the reused group, so it has to be
+                // connected by hand through the Wireless Helper.
+                AppLog.i("Existing P2P group found, delivering its credentials")
+                isGroupOwner = group.isGroupOwner
+                WifiDirectCompat.requestDeviceInfo(manager, channel) { address ->
+                    AppLog.i("WifiDirectManager: Updated localDeviceAddress via requestDeviceInfo: $address")
+                    localDeviceAddress = address
+                    manager?.requestGroupInfo(channel, this@WifiDirectManager)
+                }
+                // Fallback: if requestDeviceInfo is not supported (< API 29), call directly
+                if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
+                    manager?.requestGroupInfo(channel, this)
+                }
             }
         }
     }


o-jcardenass:
Your fix was right, I fixed a hotspot problem but didn´t made this change into wifi direct because I saw there were other branches with wifi direct fixes.

But I also found something weird while testing, not related to your fix hahaha

andrecuellar:

PR #724, opened 2026-07-29
Add a permissions step to the setup wizard

andrecuellar:
This adds a permissions step to the setup wizard.

On a fresh install the app used to fire a burst of permission prompts right after the wizard finished: MainActivity requested them blindly on launch while the wizard opened on top, and the per-feature prompts (Bluetooth, overlay, modify system settings) fired later on the home screen. It felt like the app was spamming the user.

The wizard now shows an explained permissions checklist just before the final summary. Each entry has a short help line describing what it is for, a green check when it is already granted, and an Allow button otherwise. An Enable all button requests the normal permissions in a single system dialog, and the special ones (display over other apps, modify system settings) open their own screen per row. The step is not a barrier, so the user can still move on.

The permissions come from a central AppPermissions registry, so adding a future permission is a single entry plus its two strings, with no wizard or layout changes. The same list is reachable from Settings > Permissions for users who already finished onboarding, and MainActivity now only requests permissions directly when the wizard is not going to run, which removes the overlap that caused the burst.

Covered: microphone, location, nearby devices and Bluetooth (API 31+), notifications (API 33+), display over other apps, and the optional modify system settings and legacy storage, each with the correct SDK gating. Strings are translated to all ten locales.

The onboarding version is not bumped, so existing users are not shown the wizard again; they get the new list from Settings instead.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/main/MainActivity.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/main/MainActivity.kt
@@ -31,6 +31,7 @@ import com.andrerinas.headunitrevived.aap.AapService
 import com.andrerinas.headunitrevived.app.BaseActivity
 import com.andrerinas.headunitrevived.connection.CommManager
 import com.andrerinas.headunitrevived.utils.AppLog
+import com.andrerinas.headunitrevived.utils.AppPermissions
 import android.content.res.Configuration
 import com.andrerinas.headunitrevived.utils.Settings
 import android.os.SystemClock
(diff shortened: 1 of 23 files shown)


andrecuellar:
Debug build to try this out:
https://github.com/andrecuellar/headunit-revived/releases/download/test-wizard-permissions/com.andrerinas.headunitrevived_3.2.0-beta3_debug.apk

To see the new step, either do a fresh install (or clear the app data) so the setup wizard runs, and it appears right before the final summary. If you already finished onboarding, you can open the same list from Settings > Permissions.

Heads up: uninstall your current version first. It is a debug build with a different signature, so it will not install over the Play Store one, and you will lose your settings.

andrecuellar:

PR #725, opened 2026-07-29
Fix Native AA Bluetooth handshake reliability

o-jcardenass:
## Summary

- Adds a 15s watchdog to the Native AA (Android Auto Wireless) Bluetooth handshake: after sending `WifiStartRequest`, `NativeAaHandshakeManager` called `DataInputStream.readFully()` with no timeout, so a phone that never replied left the handshake blocked forever with no error and no retry. `BluetoothSocket` has no `setSoTimeout()`, so a watchdog now force-closes the socket after 15s if no response arrives, turning a silent hang into a diagnosable failure.
- Fixes a latent bug in `NativeAaHandshakeManager.start()`: the "listener running" flag was set `true` before checking whether the Bluetooth adapter was actually ready. If that check ever failed once (e.g. at boot, before the head unit's BT stack finished initializing), the flag got stuck `true` forever, and the AA RFCOMM listener was never actually opened — every later start attempt (including manual pokes) silently no-opped instead of retrying.
- Fixes `AutoStartReceiver` starting `AapService` with a bare intent on a trusted-device Bluetooth reconnect. If the service process was already alive from an earlier session, `onCreate()`'s init never re-runs, so a Native AA handshake manager that had been stopped (e.g. after a user exit) stayed dead despite the phone reconnecting. The receiver now sends an explicit action so `onStartCommand` re-arms wireless mode only when the handshake manager isn't already active.

## Root cause context

Root-caused from user report #706, where a phone connected over Bluetooth, received `WifiStartRequest`, and never sent a response. Digging into a fresh log from the same reporter, the actual issue traced further back: no `"ACTIVELY LISTENING on Android Auto UUID"` or `"Connection accepted from"` line ever appeared in the whole session, despite two manual poke attempts that each successfully woke the phone — meaning HUR's own Bluetooth listener was never actually open. The `isRunning` flag bug above explains that exactly: a single early failure (adapter not ready yet) permanently defeats every subsequent start/retry attempt for the life of the process. The reporter tested with two different phones and got the identical result both times, pointing at HUR rather than phone-side power management.

## What this does NOT fix

This does not address why the Bluetooth adapter check might fail in the first place on a given unit (e.g. boot-order timing on some head unit chipsets). It makes that failure mode recoverable instead of a permanent, silent dead end.

## Test plan

- [x] Real hardware regression test: Unisoc-based head unit + Motorola phone — BT handshake, WiFi Direct join, SSL handshake, AA protocol negotiation, first video frame, clean disconnect.
- [x] Verified reconnect works after an in-session disconnect/exit without killing the app process (previously required a full app restart to recover).
- [ ] Awaiting confirmation from the #706 reporter.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
@@ -1708,6 +1708,24 @@ class AapService : Service(), UsbReceiver.Listener {
                     nativeAaHandshakeManager?.manualPoke(mac)
                 }
             }
+            ACTION_BT_AUTO_START          -> {
+                // AutoStartReceiver fires this on ACL_CONNECTED from a trusted device. If the
+                // service process was already alive (e.g. survived a prior disconnect/exit in
+                // this same session), onCreate()'s initWifiMode() never re-runs, so a Native AA
+                // mode that was stopped after a user exit (nativeAaHandshakeManager.stop() at
+                // disconnect) would otherwise stay dead forever despite the phone reconnecting.
+                // Only force a re-init when it's actually stopped — on a genuine cold start,
+                // onCreate() already armed everything moments ago and re-running would tear
+                // down and recreate the P2P group (new random SSID/passphrase) right as it's
+                // being delivered to the phone.
+                val settings = App.provide(this).settings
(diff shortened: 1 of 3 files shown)


andreknieriem:
Thank you very much! It is already merged ;)

PR #727, opened 2026-07-30
Restore the WiFi Launcher wake so the Wireless Helper auto-launches again (#719)

andrecuellar:
Part of #719.

This is a second, separate cause behind #719, in a different code path from PR #722 (which fixes the WiFi Direct group reuse). This one is the WiFi Helper / WiFi Launcher path, used when the phone is the WiFi host (phone hotspot).

In Helper mode the head unit wakes the phone's Wireless Helper by opening a short TCP probe to port 5289 and holding it open for a moment, so the helper notices the inbound socket and launches itself. When the two NetworkDiscovery files were consolidated into one, that brief hold was dropped, and the probe socket is now closed instantly. As a result the helper is no longer woken when it is not already running, so the user has to open the Wireless Helper by hand. Once the helper is already open the connection still works, which matches the reports.

The 5289 probe is held open for about 500 ms again before closing, restoring the wake. I also added logging around it (probe held, probe released, and a failed probe line) so a connection log clearly shows whether the wake fired.

I cannot reproduce or test this myself, since I use a wireless USB adapter and not WiFi, so this is the first time I am touching the WiFi code and I am working on assumptions. A test build and a request to confirm are on #719.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
@@ -2133,8 +2133,10 @@ class AapService : Service(), UsbReceiver.Listener {
                         }
                     }
                     5289 -> {
-                        // WiFi Launcher detected — no connection needed, just log
-                        AppLog.i("Triggered Wifi Launcher at $ip:$port.")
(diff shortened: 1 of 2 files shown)


andrecuellar:

PR #734, opened 2026-07-30
Fix Native AA connecting loop: give the quiet-host path #730's group protections (#706)

andrecuellar:
Fixes part of #706 (the Native AA "Android is connecting..." loop).

Native mode (wifiConnectionMode 3) creates its own WiFi Direct group through startNativeAaQuietHost(), which is a separate code path from the Helper mode checkGroupAndCreate() that @o-jcardenass's #730 fixed. The native path never got those protections: it only called removeGroup() before creating, never deletePersistentGroup(), and had no re-advertise or self-heal loop. So a stale persistent P2P group left over from a previous session or a reboot desyncs the phone's WPS/PBC registrar into the same endless connecting loop #730 documented for Helper mode, with nothing to recover it.

What this changes (native path only, the Helper path and #730 are left untouched):
- Purge the persistent group profile (deletePersistentGroup) before creating a fresh native group, not just removeGroup, so the recreated group does not come back with the same netId/SSID. Same idea as #730.
- The stuck PROV-DISC retry-burst self-heal now also covers native mode (it was gated to Helper only) and recreates the native group with a full purge.
- A bounded join watchdog: if no phone joins the quiet host within 30s, tear down and recreate a fresh group. After a couple of tries it drops the forced 5GHz band, for phones that cannot associate to a 5GHz group owner (a second, independent cause of a permanent connecting loop with no fallback today).
- The watchdog is cancelled and state reset on client join, disconnect, P2P off, and stop, and the recreate count is bounded so it gives up cleanly instead of looping forever.

I cannot reproduce or test this myself, since I use a wireless USB adapter and not WiFi, so this is derived from reading the code. @o-jcardenass, since it extends your #730 into the native path, your eyes would be very welcome. A test build and a request to confirm are on #706.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
@@ -734,6 +734,8 @@ class AapService : Service(), UsbReceiver.Listener {
                 AppLog.d("AapService: WiFi credentials received, but not in Native AA mode. Skipping HandshakeManager update.")
             }
         }
+        wifiDirectManager?.setNativeHandshakeStateProvider { nativeAaHandshakeManager?.isHandshakeInFlight() == true }
+        wifiDirectManager?.setNativeGroupInvalidatedListener { nativeAaHandshakeManager?.invalidateCredentials() }
 
 
         checkAlreadyConnectedUsb()
--- a/app/src/main/java/com/andrerinas/headunitrevived/connection/NativeAaHandshakeManager.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/connection/NativeAaHandshakeManager.kt
@@ -118,20 +118,23 @@ class NativeAaHandshakeManager(
         }
 
         isRunning = true
-        AppLog.i("NativeAA: Starting Bluetooth Handshake Servers...")
+        // Local Bluetooth radio address; logged on every accept so a dual-radio head unit's logs
+        // show which radio the phone actually reached (compare with the HU MAC in the phone's log).
+        val localAddr = try { adapter.address } catch (e: Exception) { "?" }
+        AppLog.i("NativeAA: Starting Bluetooth Handshake Servers (primary radio [$localAddr])...")
 
         // Start AA RFCOMM Server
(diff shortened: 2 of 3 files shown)


o-jcardenass:
Tested this on real hardware and found two issues, with a fix up at https://github.com/andrecuellar/headunit-revived/pull/1 (opened against your branch so it can merge straight in):

1. The join watchdog races the Bluetooth handshake — it has no visibility into `NativeAaHandshakeManager`, so it can tear the group down and hand out a new SSID while a handshake is mid-exchange. Reproduced this twice on-device on a phone that normally connects fine in Native mode: once it churned 4 times and never connected, once it churned once and only succeeded because the real handshake happened to land a few seconds after the recreate instead of during it.
2. `deletePersistentGroup()` is rejected outright on-device for every netId, including the profile's own real one — never actually exercised before since `netId == -1` always short-circuited it. Same call as #730's Helper-mode path, so likely affects that too.

Details, logs, and the fix are in the linked PR.

andrecuellar:
Thanks a lot for this. I pulled #1 into the branch. Your read on the watchdog racing the handshake was exactly right, and dropping the dead deletePersistentGroup call is a great catch: it explains why the purge never did anything, and it affects the helper path too.

I also saw your latest find on the issue about the Bluetooth socket being kept open after the WiFi handoff. That lines up with the logs (the phone reconnecting in a loop) and with the reference dongle project you found, so it looks like the actual root cause. Could you open that one as a PR against this branch too, the same way you did with #1? That way everything lands in one place for review.

I added a small thing on top: the native accept and handshake logs now print the local Bluetooth radio address, so on dual-radio head units we can match a HUR log against the head unit MAC the phone connected to. Happy to cut a combined test build once your Bluetooth-close fix is in.

andrecuellar:
Closing this in favor of #736, which is the complete, hardware-tested version and includes this branch's commit as-is at its base (thanks @o-jcardenass). On top of it, #736 adds the actual root-cause fix: releasing Bluetooth shortly after the WiFi handoff and no longer accepting AA connections afterward, which stops the phone's reconnect loop. Consolidating everything into #736 keeps it to one clean PR with no conflicts. I put the dual-radio diagnostic logging from here up as a small optional follow-up on @o-jcardenass's branch.

PR #735, opened 2026-07-30
Wizard and Settings UX: report-a-problem screen, quick shortcuts, live Quick Settings, animated DPI preview

andrecuellar:
This builds on the Settings redesign and the first-run wizard with a set of UX improvements, a clearer way to report problems, a fix for a crash on old Android, and a multi-select for the connection type.

Old-Android crash fix
The setup wizard crashed on launch on Android below 5.0 (reported on #647 for Android 4.4.2). Several ImageViews loaded vector drawables through android:src, which the framework cannot inflate before API 21, so the whole onboarding layout failed to inflate. They now use app:srcCompat. This also guards Display.getRealMetrics (added in API 17), which threw on Android 4.1 during the display scan, and switches the nearby-device icon to app:tint so it stays themed on old devices.

Connection multi-select
The connection choice used to be single-select (USB, WiFi, Self Mode or All). It is now a multi-select, so someone who connects more than one way can pick several and see the settings for each. Only Self Mode hides the USB, wireless and external-GPS settings; only USB hides the wireless ones; only WiFi hides the USB ones; combinations show the union. The wizard uses checkable buttons (at least one required) and Settings a multi-choice dialog. Existing choices migrate automatically, and this only changes which settings are shown, not the connection logic.

Report a problem screen
A dedicated screen with a direct link to the GitHub issues page and a QR code, so someone on a head unit with no browser or no internet can scan it from their phone. It has a short help text (search first, then open a new one) and a note that opening an issue needs a GitHub account. It is aimed at users who get lost and do not know where or how to report a bug. It also lives in the Basic tab so it is easy to find.

Loading screen in Basic
The custom loading screen setting now shows up in the Basic tab, not only in Advanced.

Wizard finishing card
The last wizard step now has a small card with quick shortcuts to the things the wizard does not cover: the loading screen, the keymap and the microphone. Each one finishes setup and opens the matching screen.

Wizard button color
The loading-screen button on the last step is now a filled, colored button so it reads as an action.

Animated DPI preview
On the DPI step the illustration animates on its own as a short demo, sweeping the size range so people understand what it does. Only the illustration moves, the slider and the number stay at the real value, and it stops the moment the user starts setting their own DPI. Same behavior in the standalone DPI screen, and it never changes the saved value.

Quick Settings additions
Two settings that apply live during projection with no reconnect: screen orientation and screen mode (system bars). Anything that needs a reconnect stays out of the overlay on purpose.

Translations
All new user-facing strings are translated across the existing locales.

A debug build to verify the old-Android crash fix is linked from #647. The branch is up to date with main and has no conflicts.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapProjectionActivity.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapProjectionActivity.kt
@@ -199,9 +199,13 @@ class AapProjectionActivity : SurfaceActivity(), IProjectionView.Callbacks, Vide
             val needsViewRecreate = intent.getBooleanExtra(QuickSettingsFragment.EXTRA_NEEDS_VIEW_RECREATE, false)
             val needsAudioRestart = intent.getBooleanExtra(QuickSettingsFragment.EXTRA_NEEDS_AUDIO_RESTART, false)
             val sensorRefresh = intent.getBooleanExtra(QuickSettingsFragment.EXTRA_SENSOR_REFRESH, false)
+            val applyFullscreen = intent.getBooleanExtra(QuickSettingsFragment.EXTRA_APPLY_FULLSCREEN, false)
 
             if (needsViewRecreate) {
                 recreateProjectionView()
+            }
+            if (applyFullscreen) {
+                setFullscreen()
             }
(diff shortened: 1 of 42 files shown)


andrecuellar:
It's ready for merge now :)

PR #736, opened 2026-07-30
Fix Native AA connecting loop and Bluetooth handshake instability (builds on #734)

o-jcardenass:
Builds on @andrecuellar's diagnosis and fix in #734 for the Native AA "Android is connecting..." loop (#706) — his commit is included as-is at the base of this branch. On top of it, testing on real hardware surfaced a few more issues in the same area, fixed here:

**1. The join watchdog raced the Bluetooth handshake.** It had no visibility into `NativeAaHandshakeManager`, so it could tear the P2P group down and hand out a new SSID while a handshake was mid-exchange. `WifiDirectManager` now checks whether a handshake is in flight before recovering and defers instead. Reproduced and fixed on-device.

**2. `deletePersistentGroup()` is rejected outright on real hardware**, for every netId, including the profile's own real one — confirmed via two independent approaches to sourcing a valid netId. Dropped the purge from both the native and Helper-mode (#730) paths in favor of the plain `removeGroup()`+`createGroup()` teardown, which is the part that's actually functional.

**3. The native join watchdog's `handshakeInFlight` flag could get stuck true permanently**, blocking recovery forever after one successful connection, because it wasn't reset until the handshake's entire "maintain session" loop exited — which can run indefinitely. Fixed by resetting it right after the credential exchange completes.

**4. HUR held the Bluetooth connection open indefinitely after the WiFi handoff succeeded.** A reference open-source wireless Android Auto dongle implementation hit the identical complaint (Bluetooth staying connected caused "confusion, especially with phone calls") and fixed it by releasing Bluetooth right after the handoff — confirmed via the real Android-side protocol behavior. Ported the same fix here: releases the connection ~3s after handoff instead of holding it forever.

**5. Just closing the socket wasn't enough** — the phone treats that as an unexpected drop and immediately retries, which the still-listening AA_UUID server would accept and bounce again in a tight loop (confirmed on-device: hundreds of accept/close cycles a second). Now stops accepting new AA_UUID connections after a successful handoff too, re-opening the next time the manager restarts.

**6. Fixed `isActive()`'s semantics** to account for the AA listener now being closeable independently of the manager's overall running state, preserving the invariant an earlier fix (re-arming on Bluetooth reconnect) depends on.

All six verified on-device across multiple repeated test runs: clean single-connect handoff, no reconnect storms, stable sessions afterward.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
@@ -734,6 +734,8 @@ class AapService : Service(), UsbReceiver.Listener {
                 AppLog.d("AapService: WiFi credentials received, but not in Native AA mode. Skipping HandshakeManager update.")
             }
         }
+        wifiDirectManager?.setNativeHandshakeStateProvider { nativeAaHandshakeManager?.isHandshakeInFlight() == true }
+        wifiDirectManager?.setNativeGroupInvalidatedListener { nativeAaHandshakeManager?.invalidateCredentials() }
 
 
         checkAlreadyConnectedUsb()
--- a/app/src/main/java/com/andrerinas/headunitrevived/connection/NativeAaHandshakeManager.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/connection/NativeAaHandshakeManager.kt
@@ -60,8 +60,14 @@ class NativeAaHandshakeManager(
     private var aaServerSocket: BluetoothServerSocket? = null
     private var hfpServerSocket: BluetoothServerSocket? = null
     // Extra RFCOMM listeners opened on secondary Bluetooth radios (dual-Bluetooth head units).
-    private val extraServerSockets = java.util.Collections.synchronizedList(mutableListOf<BluetoothServerSocket>())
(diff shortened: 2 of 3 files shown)


o-jcardenass:
Closing this in favor of routing everything through @andrecuellar's branch, per his request on the issue — the join watchdog + deletePersistentGroup fixes are already merged there (#1), and the Bluetooth-release fix is up as a new PR against the same branch: https://github.com/andrecuellar/headunit-revived/pull/2

Everything will land together in #734 for review.

andrecuellar:
Reviewed the diff, this is great and matches exactly what the logs showed. Releasing Bluetooth shortly after the WiFi handoff (and no longer accepting AA connections afterward) is the real root cause we were missing, the phone was treating the still-open socket as an unexpected drop and looping. Nicely done testing all six on real hardware.

I merged your earlier watchdog PR into my branch already, and I'm closing my #734 in favor of this so it stays one clean PR with no conflicts. I also put a tiny optional commit up on your branch that logs which local Bluetooth radio each connection used, handy for dual-radio head units like Lion62's. Take it or leave it. Thanks for driving this.

o-jcardenass:
Hahahahahaha we both closed our PRs

andrecuellar:
Haha
I will open a new PR with our fixes

PR #737, opened 2026-07-30
Fix Native AA connecting loop and Bluetooth handshake instability (#706)

andrecuellar:
Consolidates the full Native AA fix for the "Android is connecting..." loop (#706) into one PR. This replaces the two overlapping PRs we closed to avoid conflicts (#734 and #736). Every commit keeps its original author, and most of the Bluetooth-side work was found and verified on real hardware by @o-jcardenass.

What it fixes:

1. The native quiet-host path never got #730's WiFi Direct group protections, so a stale group left from a previous session or reboot kept the phone stuck connecting. (base)
2. The join watchdog raced the Bluetooth handshake and could tear the group down mid-exchange, handing the phone a new SSID. It now checks whether a handshake is in flight and defers instead. (@o-jcardenass)
3. deletePersistentGroup() is rejected on real hardware for every netId, so the purge never actually did anything. Dropped from both the native and Helper-mode (#730) paths in favor of the plain removeGroup()+createGroup() teardown that works. (@o-jcardenass)
4. The handshakeInFlight flag could stick true forever and block recovery after one connection. Reset right after the credential exchange. (@o-jcardenass)
5. The real root cause: HUR held the Bluetooth connection open indefinitely after the WiFi handoff, which the phone treats as an unexpected drop and loops on. A reference open-source wireless Android Auto dongle hit the same issue and fixed it by releasing Bluetooth right after the handoff. Ported here: releases the connection shortly after handoff and stops accepting AA connections afterward, re-opening on the next start. (@o-jcardenass)
6. Fixed isActive() to account for the AA listener now closing independently of the manager's running state. (@o-jcardenass)
7. Native accept and handshake logs now print the local Bluetooth radio, so on dual-radio head units a HUR log can be matched against the head unit MAC the phone connected to. (me)

The Bluetooth-side fixes were verified on-device by @o-jcardenass across repeated runs (clean single-connect handoff, no reconnect storms, stable sessions). Builds green.

### DIFF
--- a/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/aap/AapService.kt
@@ -734,6 +734,8 @@ class AapService : Service(), UsbReceiver.Listener {
                 AppLog.d("AapService: WiFi credentials received, but not in Native AA mode. Skipping HandshakeManager update.")
             }
         }
+        wifiDirectManager?.setNativeHandshakeStateProvider { nativeAaHandshakeManager?.isHandshakeInFlight() == true }
+        wifiDirectManager?.setNativeGroupInvalidatedListener { nativeAaHandshakeManager?.invalidateCredentials() }
 
 
         checkAlreadyConnectedUsb()
--- a/app/src/main/java/com/andrerinas/headunitrevived/connection/NativeAaHandshakeManager.kt
+++ b/app/src/main/java/com/andrerinas/headunitrevived/connection/NativeAaHandshakeManager.kt
@@ -98,7 +98,13 @@ class NativeAaHandshakeManager(
         currentBssid = null
     }
 
-    fun isActive(): Boolean = isRunning
+    // isRunning alone isn't enough once closeAaListeners() can close the AA_UUID listener while
+    // leaving the manager otherwise running (HFP stays up) — callers like AutoStartReceiver's
+    // BT-reconnect re-arm need to know whether a connection can actually be accepted right now,
+    // not just whether the manager was start()ed. See the "Re-arm on Bluetooth reconnect" fix
+    // this restores the invariant for: isActive() must mean "genuinely able to accept," not
+    // "believed to be running."
+    fun isActive(): Boolean = isRunning && !aaListenersClosedForSession
 
     fun isHandshakeInFlight(): Boolean = handshakeInFlight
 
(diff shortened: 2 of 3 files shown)


andrecuellar:
@o-jcardenass @Lion62 everything is consolidated here now: the join watchdog fixes, the Bluetooth-release-after-handoff fix and the dual-radio logging, all in one branch with authorship intact. Thanks for routing it this way, sorry about the two of us closing each other's PRs at the same second earlier.

@Lion62 if you are up for one more test, here is a debug build of this exact branch (uninstall your current version first, debug signature, will not install over the Play Store one):
https://github.com/andrecuellar/headunit-revived/releases/download/test-737-native-consolidated/com.andrerinas.headunitrevived_3.2.0-beta3_debug.apk

Whether Native AA connects or not, please export a HUR log. This build prints which local Bluetooth radio each connection used, so on your K706 / CAR8032 unit we can confirm the phone and the handshake land on the same radio. Thanks again to both of you, this was a real team effort.

andrecuellar:
On whether to merge: I think this still merits merging even though it does not resolve @Lion62's head unit. The log diagnosis in #706 shows his remaining failure is a separate issue, and it is upstream of everything this PR touches: on his dual-radio unit the phone never connects to HUR over Bluetooth at all, so the handshake never starts. The cause is a masked adapter MAC disabling the secondary-radio listener, which is unrelated to the loop fixed here.

For the common case where the phone does reach HUR, the fixes in this PR are real and were verified on hardware: the join watchdog no longer races the handshake, the dead deletePersistentGroup call is gone (it also affected the #730 helper path), and Bluetooth is now released after the WiFi handoff instead of held open, which matches the real protocol and stops the reconnect loop. With Android Auto 17.4 closing the other wireless entry points, native mode is the one we most want working.

So my vote is to merge this and track the dual-radio secondary-listener (masked-MAC) case as a follow-up, rather than close it and try to fold everything into one PR. @o-jcardenass @andreknieriem what do you both think?

o-jcardenass:
@andrecuellar opened a follow-up against your branch here: [andrecuellar/headunit-revived#3](https://github.com/andrecuellar/headunit-revived/pull/3)

Same root cause you found independently in [your comment above](https://github.com/andreknieriem/headunit-revived/issues/706#issuecomment-5136912182): `BluetoothAdapter.getAddress()` masks to `02:00:00:00:00:00` for any non-privileged app since Android 6.0, which meant the dual-radio secondary-listener path in `NativeAaHandshakeManager.start()` had never actually run on any device. Fixed by switching to the service-name identity `BluetoothHelper` already uses for the "Bluetooth Adapter" picker.

Also bundled in: continuous poke retries and an HFP-before-HSP fallback for the wake UUID (informed by two other open-source wireless AA projects), a manual secondary-radio override for future hardware where automatic detection might still miss a real second radio, and a small logging fix (the "primary radio" log lines were using the same masked address, defeating their own purpose).

All verified on-device against real hardware and unit-tested where the logic is pure enough to test (see the PR description for the full test plan). Would be glad to have it folded into this PR if it looks good to you.

https://github.com/o-jcardenass/headunit-revived/releases/download/v.3.2.0-beta3/HUR3.2.0-beta3_debug5.apk

PR #740, opened 2026-07-31
begin renaming the app to Open Headunit

andreknieriem:
As stated here: https://github.com/andreknieriem/headunit-revived/discussions/729 I need to rename the app.
Most users say Open Headunit and I think that works. So this PR is the renaming process.


o-jcardenass:
Open Headunit reads nicely. I wanted to help with the renaming and open this PR #743. I was able to test it on my phone and Self Mode works nicely. Will tests on my headunit and report back

PR #741, opened 2026-07-31
Fix dual-Bluetooth-radio detection: match by service name, not masked MAC (#706)

o-jcardenass:
> **Stacked on #743** (JNI symbol / protobuf-descriptor / rename cleanup fixes) - this branch was rebased on top of it since #743 will very likely merge into `rename-app` first. Base here is still `rename-app` (GitHub won't let a PR's base branch point at a branch that only exists on a fork), so until #743 merges, this diff will also show its commits. Once #743 lands and this gets rebased again, the diff will shrink back down to just the dual-radio changes below.

## Summary

Root cause for the dual-Bluetooth-radio head units in #706 (e.g. K706/CAR8032-style units): `BluetoothAdapter.getAddress()` returns the fixed placeholder `02:00:00:00:00:00` for any non-privileged app since Android 6.0 (API 23), on every device. The secondary-radio listener added for this class of hardware gated itself on the address being non-masked, so it never actually ran on any device, ever. Fixed by identifying radios by system service name instead - the same approach the existing "Bluetooth Adapter" picker in Settings already uses successfully.

Also bundled in:
- Continuous wake-poke retries instead of a single attempt, and trying `HFP_AG_UUID` before `HSP_AG_UUID` (mirrors two independent open-source wireless Android Auto projects that solve the same "wake the phone up" problem).
- Radio-identification log lines now print the adapter's name instead of its always-masked address, so dual-radio logs are actually useful for diagnosing which radio a connection landed on.
- A manual "Secondary Radio" override in Settings > Native, as a last-resort fallback for a unit where automatic secondary-radio detection still comes up empty - accepts either the exact system service name or the AIDL interface descriptor shown by `adb shell service list`, since both have shown up in the wild as what users copy in.

## Test plan

- [x] `./gradlew :app:testGithubDebugUnitTest` - `WifiModePolicyTest`/`NativeAaHandshakeManagerTest`/new unit tests pass, no regressions.
- [x] On-device: confirmed the secondary-radio listener actually opens and accepts a connection on real dual-radio hardware (previously dead code, confirmed via logs it never ran before this fix).
- [x] On-device: verified the poke retry loop cancels cleanly once a real handshake starts, no crash over a multi-minute idle window.
- [x] On-device: manual override tested with blank/bogus/duplicate-of-primary/real-but-wrong service names, all fail closed with no crash; confirmed zero impact on the normal single-radio path.
- [ ] Still open on #706: one dual-radio unit reaches the handshake step (Bluetooth connects, `WifiStartRequest` delivered) but never gets a response - separate, still-unresolved issue on that specific hardware, not something this PR is expected to fix.

Fixes/improves the dual-radio detection side of #706.


andreknieriem:
Hey sorry, I don't know why this happens :O Can you make the PR again?

o-jcardenass:
Hahahah sure, don't worry. It wasn't finished yet

PR #747, opened 2026-07-31
Native Mode Reliability Improvements (#706)

o-jcardenass:
Supersedes #741, closed when `rename-app` was merged and deleted out from under it - rebased onto current `main`, same branch, no content lost.

## Summary

Root cause for the dual-Bluetooth-radio head units in #706 (e.g. K706/CAR8032-style units): `BluetoothAdapter.getAddress()` returns the fixed placeholder `02:00:00:00:00:00` for any non-privileged app since Android 6.0 (API 23), on every device. The secondary-radio listener added for this class of hardware gated itself on the address being non-masked, so it never actually ran on any device, ever. Fixed by identifying radios by system service name instead - the same approach the existing "Bluetooth Adapter" picker in Settings already uses successfully.

Also bundled in:
- Continuous wake-poke retries instead of a single attempt, and trying `HFP_AG_UUID` before `HSP_AG_UUID` (mirrors two independent open-source wireless Android Auto projects that solve the same "wake the phone up" problem).
- Radio-identification log lines now print the adapter's name instead of its always-masked address, so dual-radio logs are actually useful for diagnosing which radio a connection landed on.
- A manual "Secondary Radio" override in Settings > Native, as a last-resort fallback for a unit where automatic secondary-radio detection still comes up empty - accepts either the exact system service name or the AIDL interface descriptor shown by `adb shell service list`, since both have shown up in the wild as what users copy in.
- Two targeted mitigations for the "WifiStartRequest sent, no response ever" symptom on one specific dual-radio unit, informed by a phone-side Gearhead log that finally showed what happens after the message leaves HUR: the phone's own RFCOMM read never receives a single byte before its ~12s first-message timeout fires, every time, on a stable RFCOMM channel (ruling out an SDP/wrong-channel issue). (1) Cancel the in-flight wake-poke job as soon as a real handshake starts, instead of leaving it holding a second profile channel open on the same radio for the rest of its 20s window. (2) Add a 300ms delay before sending `WifiStartRequest`, since writing immediately after accept() is a real, previously-documented class of RFCOMM race (the Linux kernel's own "Move pending packets from RFCOMM socket to TTY" fix addressed the identical symptom on the HFP profile). Neither is guaranteed to be *the* fix - the true root cause looks like it's below the app layer, in that unit's Bluetooth chip/stack - but both are safe, low-risk, and directly targeted at the mechanism the log exposed.

## Test plan

- [x] `./gradlew :app:testGithubDebugUnitTest` - all tests pass, no regressions (rebased onto main after #748 merged; builds standalone).
- [x] `./gradlew assembleGithubDebug` - builds clean.
- [x] On-device: confirmed the secondary-radio listener actually opens and accepts a connection on real dual-radio hardware (previously dead code, confirmed via logs it never ran before this fix).
- [x] On-device: verified the poke retry loop cancels cleanly once a real handshake starts, no crash over a multi-minute idle window.
- [x] On-device: manual override tested with blank/bogus/duplicate-of-primary/real-but-wrong service names, all fail closed with no crash; confirmed zero impact on the normal single-radio path.
- [x] The poke-cancel and send-delay changes are not yet re-verified live on the specific hardware they target (Lion62's unit) - next step is another round of logs from him.
- [x] Still open on #706: whether the send-delay/poke-cancel changes actually resolve that unit's "no response" symptom - the true root cause may be below the app layer entirely (this unit's Bluetooth chip/stack), in which case these are best-effort mitigations, not a guaranteed fix.

Fixes/improves the dual-radio detection side of #706.

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/connection/NativeAaHandshakeManager.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/connection/NativeAaHandshakeManager.kt
@@ -31,17 +31,13 @@ class NativeAaHandshakeManager(
     companion object {
         private val AA_UUID = UUID.fromString("4de17a00-52cb-11e6-bdf4-0800200c9a66")
         private val HFP_UUID = UUID.fromString("0000111e-0000-1000-8000-00805f9b34fb")
-        // Headset Profile Audio-Gateway role. Despite the old name this repo carried
-        // ("A2DP_SOURCE_UUID"), this is not A2DP Source (real assigned number
-        // 0000110a-0000-1000-8000-00805f9b34fb) - confirmed against two independent open-source
-        // wireless Android Auto implementations (nisargjhaveri/WirelessAndroidAutoDongle,
-        // mossyhub/openautolink), which both use this exact UUID as a phone-wake target.
-        private val HSP_AG_UUID = UUID.fromString("00001112-0000-1000-8000-00805f9b34fb")
-        // Hands-Free Profile Audio-Gateway role. openautolink's _connect_device() tries this
-        // first, falling back to HSP_AG_UUID - mirrored here for the same reason: HFP is the
-        // more modern profile and more likely what a given phone/OEM stack gates wireless AA
-        // detection on.
-        private val HFP_AG_UUID = UUID.fromString("0000111f-0000-1000-8000-00805f9b34fb")
+        // Phone-wake targets, tried HFP then HSP (mirrors openautolink's ConnectProfile
+        // fallback chain). HSP_AG_UUID is the old "A2DP_SOURCE_UUID" - despite that name it was
+        // never A2DP Source (real assigned number 0000110a-...); both UUIDs are confirmed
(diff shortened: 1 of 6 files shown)


o-jcardenass:
@andreknieriem following up on the two open checklist items here.

Got another round of logs from Lion62 after this build. The `WifiStartRequest sent, no response ever` symptom is unchanged: both mitigations (poke-cancel on handshake start, 300ms pre-send delay) are confirmed engaging correctly via the new trace logging, but his unit still times out identically every time. Compared this round's logs against the ones from before these two mitigations landed and the outcome is byte-for-byte the same either way, so neither is the fix for his specific case.

Dug into his exact hardware (QF001 / ROCO K706, Unisoc UIS7862A, non-FYT firmware) and this looks like a real hardware/vendor-stack quirk rather than anything reachable from app code. That head unit family has independent, pre-existing reports of Bluetooth firmware oddities unrelated to Android Auto or HUR at all. At this point I think his unit is a below-the-app-layer dead end for this PR, not something these changes are ever going to fix outright.

That said, I'd still like to get this merged as-is, since most of what's here is unrelated to his specific hardware and independently confirmed as real fixes

Issue #751, opened 2026-07-31
Offer a smoother migration path from Headunit Revived to Open Headunit (backup and restore bridge)


andrecuellar:
The rename changed the applicationId from `com.andrerinas.headunitrevived` to `com.andrerinas.openheadunit`. On the Play Store the applicationId is the app identity, so that change does not ship as an update. It ships as a separate listing. Existing users keep the old app installed, and Open Headunit arrives as a clean install with none of their settings (the backup format covers 100+ keys). For someone who spent time dialing in resolution, DPI, night mode, keymap and connection setup on a head unit, starting from zero is rough.

I built the rename notice modal (#746, #750) assuming a gradual handover, but as things stand the two apps are disconnected. The notice tells users about the move, yet there is no path to carry their setup across. I would like to close that gap.

Proposal: a temporary bridge where both apps coexist for a while and settings move across once.

1. Ship one more update of the old app (Headunit Revived, old package) whose notice modal does two extra things: a one tap "Save a backup" action, and a link that opens the Open Headunit store listing.
2. Write that backup to a location the new app can actually read. Because the two apps have separate sandboxes, the private app folder does not work across them. Use SAF (the document picker) or public Downloads via MediaStore so the file survives the app boundary. On Android 10+ direct Downloads file access is blocked, so SAF or MediaStore is the reliable route.
3. In Open Headunit, add a "Restore from Headunit Revived" step to the onboarding wizard that finds or imports that backup and applies it, so the first run lands the user on their previous setup.

Most of the plumbing already exists, which keeps this small:

- SettingsBackupManager already exports and imports the JSON backup and validates the format, and the import path already accepts a user picked file through SAF.
- The onboarding wizard has a clean place to add the restore step.

What is missing:

- Today the backup auto discovery only scans the app private folders, which another app cannot read, so restore needs to read from shared storage (a SAF pick, or a MediaStore Downloads query on Android 10+).
- The old app needs the export action plus a store deep link added to its notice modal.

I know the applicationId change was intentional as part of the rename, so this issue is not about reverting it. It is about softening the user impact of it. If keeping the old package was ever on the table it would avoid all of this, but assuming the new package is final, the bridge above is the least disruptive path I can see.

I am happy to implement this end to end (both the old app export side and the new app restore step) if you are on board with the approach. Open to a simpler shape if you prefer.

andrecuellar:
[...]

PR #752, opened 2026-07-31
Back up and restore all settings (fix settings missing from the backup)

andrecuellar:
The settings backup was silently skipping several settings, so restoring a backup came back incomplete. This adds the ones that were missing so export and import are complete:

- custom loading screen display options (show text, keep aspect ratio, loop video, scale)
- wireless hotspot SSID and password
- static BSSID
- the measured touch surface fix
- toast visibility

The custom loading screen media (image or video) is deliberately not backed up, because it is copied into the app private folder and its path cannot be restored on another install.

Scoped down from the earlier migration idea: since the package is staying the same, this PR is now just the backup completeness fix. Related discussion in #751.

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/utils/SettingsBackupManager.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/utils/SettingsBackupManager.kt
@@ -146,7 +146,21 @@ object SettingsBackupManager {
         "wait-for-wifi-timeout" to ValueType.INT,
         "helper-connection-strategy" to ValueType.INT,
         "bluetooth-manager-service-name" to ValueType.STRING,
-        "use-libusb" to ValueType.BOOLEAN
+        "use-libusb" to ValueType.BOOLEAN,
+        // Custom loading screen display options. The picked image/video is copied into the app's
+        // private storage, so the media path/type cannot be restored on another install and are not
+        // backed up (only these display preferences are, and apply once the media is added again).
+        "loading-screen-show-text" to ValueType.BOOLEAN,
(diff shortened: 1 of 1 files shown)


andreknieriem:
Hmm I am a little bit unsure if the package rename is mandantory. Because I lose everything from the app. Even the Playstore Listing, because I need new testers for 2 weeks etc. I think it is better to keep the package name and just rename the titles and everything which is changeable. Damn Google for not making this possible. It is even shitty for them, as hangout packagename is still com.google.talk :D 

So I will revert the package name. The normal users will never see headunit-revived anywhere besides the name in the url or in the source files. So I guess this is the best way to go. Sorry for the confusion...

andrecuellar:
Exactly, you have got it right, and no need to apologize. Keeping the package name and changing only the display title and the visible text is the painless path: it ships as a normal update, so you keep the listing, the reviews, the installs and the testers, and existing users keep all their data. Changing the applicationId was the only part that would have forced a new listing and the whole migration.

Just so the context is clear on my side: I opened the issue thinking the package change would break the old app and force a separate listing, and when I saw the package had already been changed I assumed it was deliberate and that a new APK and listing were the intended path. That is why I built the backup modal and the restore wizard, for a two app transition.

Since the package is staying, that two app migration is not needed. Here is what I am doing: I turned #752 into just the real fix that stands either way (the settings backup was silently missing some settings, like the custom loading screen options and the static BSSID). And I will open a separate PR that reverts the applicationId to com.andrerinas.headunitrevived while everything keeps displaying as Open Headunit, so the app stays the same listing and users only ever see the new name. You can keep the applicationId different from the code package, exactly like your com.google.talk example.

PR #768, opened 2026-08-03
Fix Native AA wireless handoff and handshake reliability (#760, #706)

### LINKED ISSUE #706: Native support failed on supported headunit
Hi,

[HUR_Log_20250101_010758_558.txt](https://github.com/user-attachments/files/30426909/HUR_Log_20250101_010758_558.txt)

I'm having trouble using the native mode on my car radio.

You'll find the logs below.
Thanks

### LINKED ISSUE #760: Native mode broken on 3.2.0
Something happened between 3.2.0 beta 4 and 3.2.0 official. I think its related to the new bluetooth HFP. Unpaired and repaired devices didnt fix. 

Phone: A15
Tablet: Kitkat 4.4.2

What happens: 
OHU is started, Native work flow is started. Bluetooth icon appears on phone (for 1 second, it used to appear for 5+ seconds in previous version). WiFi disconnects from router but does not connect to P2P android. I can see a network created by OHU called "DIRECT-BS-HURev" Android auto app on the phone never attempts to connect to it.

[HUR_Log_20260802_123339_437.txt](https://github.com/user-attachments/files/30636246/HUR_Log_20260802_123339_437.txt)

Edit: everyonce in a while the phone attempts to connect to DIRECT-BS-HURev" SSID and it goes through to "obtaining IP address". Every 10th try or so, it succeeds and AA connects.

o-jcardenass:
## Summary

Two Native-mode (wireless mode 3) fixes that touch the same files, so they ship together.

**#760, phone stuck on "Obtaining IP address".** After sending the phone our WiFi credentials (Type 3) we closed the Bluetooth link on a fixed 3 s timer, and the quiet host kept calling `discoverPeers()` while the phone was still joining. The phone needs longer than 3 s to associate, run WPS and get a DHCP lease (21 s in a known-good 3.1.1 log), and `discoverPeers()` takes the group owner off-channel every 10 s meanwhile. 3.1.1 worked because it never closed the link at all. Bluetooth is now held until the WiFi session reports up, bounded at 45 s, and the group owner stays on-channel on the native path.

**#706, log looks healthy but the phone never connects.** The wait for the phone's Type 2 was bounded by a watchdog that closed the socket to interrupt `readFully()`. That only works where `close()` interrupts a pending read, and the reporter's stack does not, so `handleHandshake()` never unwound: handshake state stayed latched for the life of the process, the wake poke stopped retrying, and the P2P join watchdog deferred recovery forever. The read now runs under `withTimeoutOrNull` and resumes on schedule whether or not it ever returns.

## Changes

| File | What |
|---|---|
| `aap/NativeHandoffPolicy.kt` (new) | Pure policy object with the timing rules both managers need: `isSettling`, `isHandshaking`, `shouldPoke`, `shouldWarnPhoneNeverCallsBack`, `shouldServeHandshake`, `shouldRestartDiscovery`. The two windows are bounded stamps, not booleans, so a coroutine that never unwinds cannot latch them true. |
| `connection/NativeAaHandshakeManager.kt` | Hold Bluetooth until the WiFi session lands instead of 3 s; run the Type 2 read under `withTimeoutOrNull` and unwind on timeout; stop serving handshakes after 5 consecutive timeouts; warn when the phone answers pokes but never opens the AA channel; stop `pokeDevice()` swallowing cancellation and firing a second connect. |
| `connection/WifiDirectManager.kt` | Never restart peer discovery on the native quiet-host path; sync auto-start-on-boot/USB to device-protected storage when disabled. |
| `aap/AapService.kt` | Stop our own HFP poke self-triggering a Bluetooth auto-start re-arm; stop the native join watchdog tearing down live sessions. |
| `main/AutoStartFragment.kt` | Settings write path for the auto-start sync. |
| `test/.../NativeHandoffPolicyTest.kt` (new) | 22 JVM unit tests covering every rule in the policy object. |

## Worth a closer look

`shouldServeHandshake` is the only hard stop here. The Type 2 wait is a blocking read with no suspension point, so where `close()` does not interrupt it the coroutine's `Dispatchers.IO` thread is stranded for the life of the process. That pool caps at 64 and the reporter's phone reconnects every ~12.8 s, so unbounded retrying exhausts it in about 15 minutes. Only timeouts increment the counter; a reader that returns clears it, and so do `stop()` and a manual poke, so the UI's "try again" is always a way back.

## Verification

CI green on the branch head: **Build (github debug)** and **Unit tests (github debug)** both pass.

Confirmed on-device by the #760 reporter, on the hardware that showed the fault:

- 5/5 successful connections with the #760 fixes: https://github.com/andreknieriem/open-headunit/issues/760#issuecomment-5182718700
- 8/8 with the #706 commits added, 5 started from the app and 3 by connecting Bluetooth manually from the phone: https://github.com/andreknieriem/open-headunit/issues/760#issuecomment-5186305558

## Out of scope

This does not fix #706's root cause. On that head unit the RFCOMM write never reaches the air (an HCI trace shows 15 complete sessions and not one data frame), and nothing in this app can make the OEM Bluetooth module answer. These commits bound what the failure costs and name the likely cause in the log.

A failed handoff now takes up to 45 s to fall back to a poke retry instead of 3 s. That is the intended trade: the 3 s timer is what broke #760.

APK: https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.0_debug.apk

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/AapService.kt
@@ -735,7 +735,13 @@ class AapService : Service(), UsbReceiver.Listener {
                 AppLog.d("AapService: WiFi credentials received, but not in Native AA mode. Skipping HandshakeManager update.")
             }
         }
-        wifiDirectManager?.setNativeHandshakeStateProvider { nativeAaHandshakeManager?.isHandshakeInFlight() == true }
+        // Settling counts as in-flight here: isHandshakeInFlight() goes false the instant Type 3
+        // is written, but the phone still has to associate, do WPS and get a DHCP lease, and
+        // recreating the group in that window hands it an SSID it can no longer join.
+        wifiDirectManager?.setNativeHandshakeStateProvider {
+            nativeAaHandshakeManager?.isHandshakeInFlight() == true ||
+                nativeAaHandshakeManager?.isHandoffSettling() == true
+        }
         wifiDirectManager?.setNativeSessionConnectedProvider { commManager.isConnected }
         wifiDirectManager?.setNativeGroupInvalidatedListener { nativeAaHandshakeManager?.invalidateCredentials() }
 
@@ -1738,8 +1744,14 @@ class AapService : Service(), UsbReceiver.Listener {
                 // onCreate() already armed everything moments ago and re-running would tear
                 // down and recreate the P2P group (new random SSID/passphrase) right as it's
                 // being delivered to the phone.
+                // A successful handoff closes the AA listeners, so isActive() is false for the
+                // whole life of a working session — without the connection check below, any
+                // later ACL_CONNECTED (the phone's own Bluetooth profiles reconnecting, or one
+                // of our pokes) would tear down a session that is projecting fine.
                 val settings = App.provide(this).settings
-                if (settings.wifiConnectionMode == 3 &&
+                val sessionUp = commManager.isConnected ||
+                    commManager.connectionState.value is CommManager.ConnectionState.Connecting
+                if (settings.wifiConnectionMode == 3 && !sessionUp &&
                     nativeAaHandshakeManager?.isActive() != true &&
                     nativeAaHandshakeManager?.isAttemptInFlight() != true) {
                     AppLog.i("AapService: Bluetooth auto-start — Native AA handshake manager was stopped, re-arming.")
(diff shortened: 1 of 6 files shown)


andreknieriem:
This is still in testing? Because we should bring 3.2.1 fast enough to the users ;)

o-jcardenass:
Got it, I will revert one commit so we make native usable, not reliable, but usable

o-jcardenass:
Tested on #760

PR #769, opened 2026-08-03
Refactor wifi-code from AapService into their own classes

MrEAlderson:
The purpose of these rather major changes is purely to improve maintenance, since components related to wireless connection are scattered throughout (the already large) AapService class. More specifically, this PR structures each mode into its own designated class, implements a single manager that ensures that no component/service is active when it shouldn't, and utilize enums instead of rather-cryptic ints for wifi mode and strategy. About 400 lines have been pulled out of AapService.

Do be noted that there is the potential for things to break, although I haven't faced any issues so far.

Further:
- Fixes a bug where clients aren't able to connect again for 5s (due to built-in cooldown), even though it has been clicked on within the main menu. As a result, one would get stuck within "Android auto is starting"


o-jcardenass:
This is great, I already find it difficult to trace problems and this will surely help a lot.

I suggest to move on with this PR after we fix the current 3.2.0 issues to avoid conflicts.

andreknieriem:
This is a really good idea! The CommManager was the first big think refactored from the AapServive. This is a logical and good step!

andreknieriem:
@o-jcardenass and @MrEAlderson You are both working right now on changes to the aapservice. So keep in mind, that this will likely cause merge conflicts ;)

MrEAlderson:
> @o-jcardenass and @MrEAlderson You are both working right now on changes to the aapservice. So keep in mind, that this will likely cause merge conflicts ;)

Gonna pull any changes from master and fix the conflicts, no worries about that

andreknieriem:
@MrEAlderson, can you please check this? I plan to merge this for the new 3.3.0 version. The 3.2.x are bugfix releases without ne features ;)

MrEAlderson:
> @MrEAlderson, can you please check this? I plan to merge this for the new 3.3.0 version. The 3.2.x are bugfix releases without ne features ;)

Thanks for the heads up! Fixed all the conflicts

Issue #767, opened 2026-08-03
3.2.0 stuck at Android Auto is starting..., after using the new setup wizard


cbwee:
When the app was updated from 3.2.0-alpha to the 3.2.0 release, the Android Auto screen could be displayed correctly, but the app showed the setup wizard later, and after going through the setup, the AA screen is always stuck at "Android Auto is starting...", while the audio still works correctly for Android auto at the same time.

Please see the logs, for before using the setup wizard and after, in the attached files.
[HUR_Log_20260803_151855_133_before.txt](https://github.com/user-attachments/files/30664863/HUR_Log_20260803_151855_133_before.txt)
[HUR_Log_20260803_152209_703_after.txt](https://github.com/user-attachments/files/30664864/HUR_Log_20260803_152209_703_after.txt)

My headunit has MT6735 CPU, it runs Android 5.1.

andreknieriem:
The logs say it very clear. Video Codec h264 before and h265 (hvec) after. That it just wrong. Can you please look at the settings and change the video codec back to h265

cbwee:
Hi, I just went to the car and tried using both H264 and H265, and both of them all don't have the correct behaviours (the app was working correctly, before running this new setup wizard which changed the settings, the 133_before log was the correct working one with the old settings):

Now below are the behaviours and logs with the new settings changed by the setup wizard with H264 and H265:

- H264 - it could display but kept flickering:

https://github.com/user-attachments/assets/4b5b0c1e-4bc9-47df-b356-f31850eda753

[HUR_Log_20260803_223955_407_H264.txt](https://github.com/user-attachments/files/30667012/HUR_Log_20260803_223955_407_H264.txt)


- H265 - it was completely stuck at any of these screens (but later it's stuck at Android Auto is starting... again):
<p align="center">
  <img src="https://github.com/user-attachments/assets/472a9f77-827e-45a4-a6bc-819064cc7b09" width="25%" />
  <img src="https://github.com/user-attachments/assets/20569e37-ab10-45f2-8fc1-0c6e128fb59a" width="25%" />
  <img src="https://github.com/user-attachments/assets/31077578-a4c2-4613-bc8c-e8e918e6e1d6" width="25%" />
</p>

[HUR_Log_20260803_224303_370_H265.txt](https://github.com/user-attachments/files/30667118/HUR_Log_20260803_224303_370_H265.txt)
[...]

PR #775, opened 2026-08-04
Fix the setup wizard breaking the display, and make DPI and resolution consistent

andrecuellar:
Some head units ended up on a black Android Auto screen (audio still working) after going through the new setup wizard (#767). I traced it down: the black screen was the renderer, not the DPI or the resolution. The wizard was writing SurfaceView on devices where SurfaceView does not draw, and since a broken SurfaceView reports no frames, the stall watchdog from #650 could not recover it.

While I was in there I also fixed the DPI calculation and unified the resolution rules, which is where the confusing "recommended values look wrong" reports were coming from.

## Wizard
- It no longer overwrites the display config (view mode, codec, resolution, orientation) for people who already had the app working. Only fresh installs get the recommendation.
- Fresh installs now start on TextureView instead of SurfaceView on 5.0 to 8.1 with HEVC. TextureView failures are observable, so the existing stall watchdog can still escalate to SurfaceView on its own if a device needs it.
- After the wizard changes the renderer, the first projection shows a small "Do you see the screen?" bar with a "Switch renderer" action that rotates the backend live. It also appears when the phone is streaming video but nothing is drawn, and when the watchdog escalates to SurfaceView, so a blank renderer that cannot be detected in software is no longer a dead end.

## DPI
- The recommended DPI was computed from the raw panel pixels. When the picture is downscaled that over-reports the density. It is now computed from the resolution Android Auto actually draws (the panel-capped one).
- The old 1.2 multiplier is now a named, documented legibility factor set to 1.1.

## Resolution
- There were three different definitions of "what fits the panel" (the too-high warning, the runtime cap and the recommendation) and they disagreed at the edges, so the warning could say one thing and the engine do another. They are unified into a single panel-ceiling source of truth, so the warning, the cap and the DPI always agree.
- When a chosen resolution is above the panel, the image is still limited to what the panel supports (hard cap, unchanged), and the dialog now says so honestly instead of implying the higher value will be used.
- The DPI screen and the wizard DPI step now warn, in red, when the chosen resolution is above the panel and the DPI is computed for the capped one.

Also cached the hardware HEVC check since it scans the whole codec list and is now used on UI paths, and updated the translations for all locales.

Related to #767 and #715.

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/AapProjectionActivity.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/AapProjectionActivity.kt
@@ -15,6 +15,7 @@ import android.view.TextureView
 import android.view.View
 import android.widget.Button
 import android.widget.FrameLayout
+import android.widget.LinearLayout
 import android.widget.TextView
 import android.widget.Toast
 import androidx.activity.OnBackPressedCallback
@@ -137,7 +138,9 @@ class AapProjectionActivity : SurfaceActivity(), IProjectionView.Callbacks, Vide
             }
             val lastFrame = videoDecoder.lastFrameRenderedMs
             if (lastFrame == 0L) {
-                // First frame hasn't arrived yet — handled by the starting overlay
+                // First frame hasn't arrived yet — handled by the starting overlay. If the phone is
+                // streaming video but nothing draws, offer to switch renderer (issue #767).
+                maybeOfferRendererConfirm()
                 watchdogHandler.postDelayed(this, 2000)
                 return
             }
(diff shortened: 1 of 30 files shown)


andreknieriem:
@andrecuellar I've already merged it. But h265 should only be presented if res > full hd. There are soo many errors with faulty h265 devices which lie, that they can handle it, so a h264 as default is the better way, or what do you think?

andrecuellar:
I did not know that, I had to look into it. My code was asking the head unit directly whether it supported H.265 and trusting the answer, I did not know that many of them lie about it, they say they support it and then do not handle it well. So you are right, H.264 as the default is the safer call. I made the wizard default to H.264 and only pick H.265 when the panel is above Full HD (1440p or 4K), where H.264 bandwidth stops being practical. The runtime already caps the resolution to the panel, so it lines up. I opened a small follow-up for it: #782.

Issue #774, opened 2026-08-04
Native connection crashing on loop


Dalamy97:
Hello,

I've been using this app successfully with release 3.1 and 3.2, using the wireless helper and with the hotspot created by the headunit.
I choose this type of connection because the native one was giving me problem (the crashes I'll report below).
I'm still on AA 17.2 and i'll stick on this version since I saw that 17.4 is breaking things with the wireless helper. This gave me the will to debug the problem I had with native connection.
Even if I've the HU hotspot+wirelss helper working, I would like to have the native one also to work. Or at least to try and understand why I'm having crashes.

I've made a video to make clearer what I see. Imagine that what I see is this video on loop: open-headunit opens, the phone connects with the HU, android auto appears on the HU (sometimes it doesn't make in time to appear and crashes), the HU crashes, and the cycle repeats.

Video:

https://github.com/user-attachments/assets/e00617ec-94be-4908-ae46-f8b2db46fba6

[HUR_Log_20260804_210151_062_090301.txt](https://github.com/user-attachments/files/30718142/HUR_Log_20260804_210151_062_090301.txt)

Thank you for your help!

o-jcardenass:
EDIT: I did checked your log

One good piece of news: your Native AA handshake itself works perfectly. Nothing wrong there, so don't worry about the other issues on Native AA.

But, I think your unit can have a fragile Wi-Fi Direct

Could you try, Wireless Helper mode with connection method "WiFi Direct"? 
That uses the same P2P group without any of Native AA. If it still reboots, Wi-Fi Direct alone is the problem.

Dalamy97:
Hello,

I tried with Wireless Helper mode with connection method "WiFi Direct". Below the log.
It seemed to work fine, I just kept it connected for a minute or so, as it seemed to work correctly whereas with native it would crash in the first few seconds. Tell me if you need a log with a connection for longer time. 

[HUR_Log_20260805_080953_560_081035.txt](https://github.com/user-attachments/files/30732074/HUR_Log_20260805_080953_560_081035.txt)

o-jcardenass:
@Dalamy97 please try Native mode with the newest version v.3.2.1 and share me the logs again. Make sure to change the log level to Debug
https://github.com/andreknieriem/open-headunit/releases/download/v.3.2.1/com.andrerinas.headunitrevived_3.2.1.apk

andreknieriem:
[...]

PR #785, opened 2026-08-06
Experimental Feat - Native AA Hotspot route

o-jcardenass:
## What this adds

A second transport for Native AA wireless (mode 3). Instead of hosting a WiFi Direct P2P group, the head unit hands the phone the credentials of its own hotspot, and the phone joins it like any other network. Same Bluetooth handshake, same UUID, no P2P anywhere in the path.

The route is selected in Android Auto mode → Transport, and defaults to WiFi Direct. Nothing changes for anyone who does not switch it.

## Why

Most Native AA bug reports are P2P failures rather than protocol failures: group churn, the self-wake loop, phones stuck on "Obtaining IP address". A teardown of the OEM ZLink app showed it doing wireless Android Auto over an ordinary WPA2 soft AP, which proves the phone accepts a plain access point and that the P2P group is a choice rather than a requirement.

It also turns out to be the fix for a specific class of hardware failure. In #774 the reporter's MT8163 unit reboots its whole Android whenever a phone joins a WiFi Direct group the app created. That unit cannot use Native AA over WiFi Direct at all. Over its own hotspot, it never touches the failing path.

## How it works

SoftApCredentialsProvider resolves the access point the device is already running: SSID and passphrase from the system hotspot configuration (with manual overrides for devices that refuse the read), the interface by elimination against the station address, and the BSSID from that interface. It watches the AP state and invalidates the credentials if it goes down. AapService routes mode 3 to it instead of WifiDirectManager when the transport is set to Hotspot, and both transports feed the same credential listener so the poke rules stay in one place.

## Also in here

The handshake is a state machine now. WppHandshakeSession owns the ordering rules and is pure; NativeAaHandshakeManager does only I/O. What this replaces sent type 1, read one message, sent type 3 and stopped listening, so types 6 and 7 (the phone's own join feedback) were never seen. That feedback is the only join signal the hotspot route can have, since a SoftAP has no equivalent of the P2P client list. It also means the AA listeners now close when the phone's session actually lands rather than on a blind 3 s timer.

The BSSID is required, and we now always send one. Our wireless.proto had relaxed bssid, security_mode and access_point_type to optional on the strength of a claim that other implementations ship without them. That claim was wrong: aa-proxy-rs and WirelessAndroidAutoDongle both declare them required and read the address from the AP interface. Measured against a current Gearhead 17.4, credentials with no BSSID are refused outright with WIFI_INVALID_BSSID. The fields stay optional locally so a truncated message still parses.

A boot crash-loop guard. Auto-start on boot turns one system-level crash into an endless cycle, measured at a 20–26 s period across seven consecutive processes. After three boot-started runs that fail to last 30 s, wireless bring-up is skipped and a notification says why. USB is unaffected, and opening the app turns it back on.

Two diagnostics for failures that previously looked like nothing. A session that reaches SSL and ends having rendered no frames, three times running, now says so and names both ways out, that shape is what a link too slow to carry the stream looks like. And a missing BSSID now explains itself instead of failing silently.

## Testing

11 runs on two devices across three rounds, on a dedicated head unit and phone-to-phone. WiFi Direct regression runs pass, so the existing route is unaffected.

The 2.4 GHz behaviour is worth knowing about: at 1080p/60 the phone connects, opens the video channel and drops the session repeatedly without a single frame arriving. On the same access point at 800x480/30 it held indefinitely, 66,137 decoded frames. It is a throughput ceiling, not a broken band. 5 GHz at full quality is clean.

## Risk

The hotspot route is inert unless selected. The one change that alters the wire for existing mode 3 users would have been the version exchange (type 4), so it is off by default, the version it announces is a guess, no capture of a real head unit's type 4 has been decoded, and turning it off is byte-identical to 3.2.1. It stays available as an opt-in for anyone whose unit broke on Android Auto 17.4, which is the only reason to send it.

Everything else is either additive, log-only, or behind the transport setting. The starvation counter is the one piece that runs on every transport including USB; it reads a frame count at disconnect and can only produce a log line.

## New Possibilites

Native now can work Phone-to-Phone, but a phone acting as head unit cannot read its own MAC, as Android hides it, it's required to get them via adb commands or a rooted phone.
Native WiFi Direct only requires manually setting BSSID and
(New) Native Hotspot requires BSSID, SSID and PASSWORD

1. Read the address (on the headunit phone)
Hotspot transport: 

The AP must be up first, the address doesn't exist otherwise:
```
adb shell cmd wifi start-softap OHU-TEST wpa2 testtest1234 -b 5
adb shell dumpsys wifi | grep -i SoftApInfo
# SoftApInfo{bssid=ff:ff:ff:ff:ff:ff, frequency=5240, bandwidth=2, ...}
```
```-b 5```  sets the band on 5 GHz which enables 1080p+ resolutions, without it the hotspot creates as 2.4 GHz

WiFi Direct transport:
```
adb shell dumpsys wifip2p | grep -A 3 mThisDevice
#  deviceAddress: aa:aa:aa:aa:aa:aa
```
2. Write the settings in the manual fields depending on the wireless method selected for Native
3. The two phones must be paired

## Known limitations

- Some devices regenerate their hotspot BSSID on every restart, which makes the static BSSID override unusable for this route on that hardware.
- The band cannot be read without NETWORK_SETTINGS, so the 2.4 GHz problem is diagnosed by symptom rather than detected.
- A device that will not let an app read its hotspot configuration needs the SSID and password entering by hand.

## APK
https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/Experimental_OHU3.2.2_debug.apk

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/AapService.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/AapService.kt
@@ -722,7 +722,14 @@ class AapService : Service(), UsbReceiver.Listener {
             }
         }
 
-        initWifiModeWithOptionalWait()
+        // Decided here as well as inside initWifiMode() so a paused start skips the wait-for-WiFi
+        // machinery entirely rather than setting it up and being turned away at the end of it.
+        if (applyBootLoopGuard()) {
+            AppLog.w("AapService: Wireless bring-up paused by the boot-loop guard. USB and the rest of the app are unaffected.")
+        } else {
+            initWifiModeWithOptionalWait()
+        }
+        scheduleBootLoopStrikeClear()
         wifiDirectManager?.setCredentialsListener { ssid, psk, ip, bssid ->
             onNativeCredentials(ssid, psk, ip, bssid)
         }
@@ -1433,6 +1440,15 @@ class AapService : Service(), UsbReceiver.Listener {
 
     /** Starts [WirelessServer] if the user has configured server WiFi mode. */
     private fun initWifiMode(force: Boolean = false) {
+        // Every automatic entry point lands here, including the Bluetooth auto-start that fires
+        // when the phone comes into range — which on a looping unit would walk straight back into
+        // the crash the guard was set to avoid. Explicit user actions release the pause first, so
+        // this only ever blocks a start nobody asked for.
+        if (Settings.isWirelessPausedByBootLoop(this)) {
+            AppLog.w("AapService: Wireless bring-up requested, but it is paused by the boot-loop guard. Open the app to re-enable it.")
+            return
+        }
+
         val settings = App.provide(this).settings
         val mode = settings.wifiConnectionMode
         val strategy = settings.helperConnectionStrategy
@@ -1702,7 +1718,12 @@ class AapService : Service(), UsbReceiver.Listener {
 
         when (intent?.action) {
             ACTION_START_SELF_MODE       -> startSelfMode()
-            ACTION_START_WIRELESS        -> initWifiMode()
+            ACTION_START_WIRELESS        -> {
+                // Asked for from the UI, so the user is present: release the boot-loop pause
+                // rather than silently ignoring them.
+                Settings.clearBootLoopState(this)
+                initWifiMode()
+            }
             ACTION_START_WIRELESS_SCAN   -> {
                 val settings = App.provide(this).settings
                 val mode = settings.wifiConnectionMode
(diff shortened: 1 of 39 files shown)


andreknieriem:
Wow! This is absolutely insane! I will check this out. 3.2.2 will be released soon with the bugfixes. Do you think I should merge this before or after?

o-jcardenass:
You can do it after if you want, to ensure more people can test it. It shouldn't break anything by his own because that hotspot route is isolated but who knows hahahaha. 

The only PR that worries me is this one https://github.com/andreknieriem/open-headunit/pull/769 as it moves all and can be messier to fix hahaha

andreknieriem:
Haha, yes this rewrite will be in a next major release like 3.3.0 together with my theming options. First bugfixing ;)

Issue #755, opened 2026-08-01
Open HU video blinking


AyazIsayev:
There is video blinking periodically. Like different frequency ranges on monitors issue. Sound is ok.

[HUR_Log_20260801_153455_237.txt](https://github.com/user-attachments/files/30621566/HUR_Log_20260801_153455_237.txt)

[WirelessHelper_Log_20260801_153533.txt](https://github.com/user-attachments/files/30621568/WirelessHelper_Log_20260801_153533.txt)

notathf:
It also flickers when the picture is still. Apparently the decoder thinks that the connection is lost and needs to be refreshed

o-jcardenass:
Please try this, and tell me if it works:
https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.0_debug.apk

AyazIsayev:
Tested. The flickering remains, but not as often as before.

[HUR_Log_20260803_084155_411.txt](https://github.com/user-attachments/files/30647033/HUR_Log_20260803_084155_411.txt)

[HUR_Log_20260803_084146_785.txt](https://github.com/user-attachments/files/30647035/HUR_Log_20260803_084146_785.txt)

[WirelessHelper_Log_20260803_084253.txt](https://github.com/user-attachments/files/30647036/WirelessHelper_Log_20260803_084253.txt)

AyazIsayev:
Another logs.

[HUR_Log_20260803_132809_431.txt](https://github.com/user-attachments/files/30656341/HUR_Log_20260803_132809_431.txt)

[HUR_Log_20260803_132519_193.txt](https://github.com/user-attachments/files/30656351/HUR_Log_20260803_132519_193.txt)

[HUR_Log_20260803_133120_265.txt](https://github.com/user-attachments/files/30656360/HUR_Log_20260803_133120_265.txt)

o-jcardenass:
Thanks for the logs, could you please retry with this. (same link as before, but updated apk)
https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.0_debug.apk

AyazIsayev:
> Thanks for the logs, could you please retry with this. (same link as before, but updated apk) https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.0_debug.apk

Tested. No flickering already, but the video is freezing like it is on 15-20 FPS.

[...]

PR #789, opened 2026-08-06
Recover a dropped video frame without releasing video focus

andrecuellar:
On some head units the video collapses to a few fps and freezes shortly after connecting, and the last version that worked for those users is 3.2.0-beta2 (issue #755).

From @AyazIsayev's logs (clean install, H.264): the video runs fine at about 48 fps, then a transient decoder input-buffer-full drops one frame. The stream gets marked corrupt and the recovery asks for a keyframe by cycling video focus (release, then re-gain). Right after that, the phone drops to about 3 fps and never recovers, and the session ends with a "Magic Garbage detected in header" desync. beta2 does not have this recovery path, which is why it stays at full framerate there (a dropped frame just self-heals on the next keyframe).

The change: request the recovery keyframe with an unsolicited focus gain only, without releasing focus first. Releasing focus (gain=false) is what tells the phone to stop sending video, and on these units it does not come back to full rate. A gain-only request is the same nudge the startup keyframe watchdog already uses, and it still makes the phone resend a keyframe without disturbing the running stream.

This does not touch the genuine-corruption paths (fragment overflow) or the sync_stall watchdog, only how the keyframe is requested. @o-jcardenass this sits right on top of your recovery and buffer-full work, so I would appreciate your eyes on it.

I cannot reproduce this myself (I use a USB adapter, not WiFi), so it is based on the logs. A debug build is on the issue for @AyazIsayev to confirm.

Related to #755.

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/AapTransport.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/AapTransport.kt
@@ -129,13 +129,14 @@ class AapTransport(
         get() = pollThread?.isAlive ?: false
 
     private fun triggerFocusCycleRecovery() {
-        AppLog.w("AapTransport: Requesting recovery keyframe via focus cycle.")
-        send(com.andrerinas.openheadunit.aap.protocol.messages.VideoFocusEvent(gain = false, unsolicited = false))
-        android.os.Handler(android.os.Looper.getMainLooper()).postDelayed({
-            if (isAlive) {
-                send(com.andrerinas.openheadunit.aap.protocol.messages.VideoFocusEvent(gain = true, unsolicited = true))
-            }
-        }, 100)
+        // Ask the phone for a fresh keyframe with an unsolicited focus GAIN only, the same nudge the
+        // startup keyframe watchdog uses. We deliberately do not release focus (gain=false) first:
+        // on some head unit / phone combos, releasing video focus drops the stream to a few fps and
+        // it never recovers, so a single transient dropped frame turned into a permanent freeze
+        // (issue #755). A gain-only request still makes the phone resend a keyframe without
+        // disturbing the running stream.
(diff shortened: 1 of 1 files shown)


andreknieriem:
The question is, why does not happen before. There must be a bug. I will maybe look at this too. Maybe its the stall detector which takes to much cpu or something?

o-jcardenass:
I was also tracing it because just after the 3.2.0beta4 my headunit started flickering that is why I put my fixes, maybe it comes from this PR #738 which has changes on AapVideo and VideoDecoder, and so far the tendency is that versions prior to that work

andreknieriem:
uhh #738 maybe the cause. I am not at my pc right now. If you want, can you reset the files to the version before and let the users test that apk?

o-jcardenass:
I ran this test on my revert branch
PASS Build + unit tests, both branches
PASS 3 min 1080p/H.264 A/B vs baseline: 45 samples, fed == rendered, dropped=0, 49-60 fps
PASS 30 min soak (music + nav): 368 samples, dropped=0, zero sync_stall, 58.2 fps avg, no drift
PASS 3 min explicit H.265: 50 samples, dropped=0 (buffer growth never needed at 1080p)
PASS Ping cadence at light load: 250 samples, median 1.0 s, max 1.111 s; audio continuous
PASS Nine decoder restarts: all recovered clean, no smearing
NOT RUN Stress under load: unit hard-reboots (uptime 62 s), 5 attempts, identical on both builds
NOT RUN Codec-mismatch latch: needs a 2560x1440 panel, this one caps to 1080p
NOT HIT Smearing probe: target window never reached across the nine restarts

PR https://github.com/andreknieriem/open-headunit/pull/792 
I didn't make any change to the settings wizard nor the dpi

Give it a try https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.3_videorevert.apk

Issue #788, opened 2026-08-06
Only one specific version works


mrkontrast-coder:
So for My divace only the 3.2.0-beta3_debug version works best, with 60 fps and wifi direct working best, only the media playback stops every time I lock or unlock the phone for some reason. 

All the recent version either don't work with wifi direct or have 4 fps.
Don't know if that helps 😅

mrkontrast-coder:
Android Auto version 17.2.662634-release

andreknieriem:
Take a look at this issue: 
https://github.com/andreknieriem/open-headunit/issues/755#issuecomment-5207484755

And try this apk.

Issue #787, opened 2026-08-06
3.2.2 Installed. Hello of lag on all apps.


mani169mani:
With the recent update, my maps, music and various other apps are lagging. Not able to use. Very slow response. What's the fix. It was working so smooth on 3.1.1

o-jcardenass:
Hi, please share the logs so we can take a look.

Settings -> Advanced -> Change Log Level to Debug -> Start Capturing Logs
Go back and use the app as you normally do, and then go back to
Settings -> Advanced -> Export Logs
and reply here with the log file attached

Farruco416:
These are the same symptoms I've observed and already mentioned in post #744. Since version 3.2, the app has experienced worse performance.

andreknieriem:
Please give this a try: https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.3_videorevert.apk

Issue #771, opened 2026-08-03
Can't connect with any method to 3.2.0; reverting to 3.1.1 still works


CSX321:
I didn't see quite my exact issue in either Issues or Discussions.
I'm running HR/OH on my old Pixel 6 in my car, connecting from my Pixel 9 Pro. The P6 is rooted and configured to turn on when it detects power at the USB port. HR/OH is set to start at boot, which it does. The P6 is running a script to monitor the USB connection, and it powers off 20 seconds after I shut the car off. This has all been working great for over a year.
Connection mode is Google Nearby (Beta) using Wireless Helper on my P9P. AA version is still 17.2.
With 3.1.1 on the P6 this was working most of the time, with some issues. With 3.2.0 on the P6, I have not been able to successfully connect at all with any settings or connection options. Reverting to 3.1.1 on the P6 restores the previous behavior.
With 3.1.1 it seems that about 20% of the time everything works normally and the connection happens automatically.
About 70% of the time, the connection is never made no matter how long I wait, and the only button on HR/OH that responds is Exit. To connect I have to hit Exit, relaunch HR/OH, hit Wifi, and tap my phone's name. Then it connects and works normally.
The remaining 10% of the time, I hit Exit as above, but when I relaunch and hit Wifi, the list is empty. When this happens, I have to force close HR/OH on the P6, force close Wireless Helper on my P9P, and relaunch both. Normally that allows me to follow the 70% scenario and connect.
In any case, Google Nearby seems completely broken for my configuration in 3.2.0, but still works in 3.1.1 with the same Wireless Helper version and same everything else.
If I export logs, do you need both sides (Wireles Helper and HR/OH)?

o-jcardenass:
Hi, which Wireless Helper version are you using? 

Alongside the changes of OHU 3.2.0 it was also necessary to change WH, which now sits at version 1.9.3
I have tested OHU 3.2.0 + WH 1.9.3 + AA 17.2 and it works

CSX321:
Ah, that makes sense. Only 1.9.2 is showing up in the Play Store for me, but I installed 1.9.3 from github and indeed it works with 3.2.0. Thanks.

andreknieriem:
Oh wow, thanks for mention that. Playstore has limited the Release to 20% and not 100% rollout. Never seen that. I've changed that too 100%.

Should now be available for all.

Issue #757, opened 2026-08-02
Version 3.2.0 broke HW decoding on KitKat


zd3sf:
Something happened between 3.2.0 beta 4 and the official stable. 

HW decoding no longer works. stuck on "Android Auto is starting"

The logs show something along the lines "SW decoding: null, HW decoding: null, preferHW: yes, selected: null"

Tablet: Dell Venue 3730, CPU: Intel Atom Z2560 (x86)

o-jcardenass:
Hi, there was only one change in video decoder intended to fix a bug with H.264 that beta4 introduced #749. I can only assume that you are running H.265 in a unit that don't supports it.

But to be sure, please share us both logs. One with a run in beta4 and other with 3.2.0.

That would help us find the real problem

zd3sf:
Im definitely on H.264. 

[HUR_Log_20260802_055504_427.txt](https://github.com/user-attachments/files/30630866/HUR_Log_20260802_055504_427.txt)

<img width="128" height="80" alt="Image" src="https://github.com/user-attachments/assets/ca7632b0-df92-430a-b4ff-a63207a2ce9e" />

o-jcardenass:
Please try this, and tell me if it works:
https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.0_debug.apk

zd3sf:
Yes this version worked with no problems. Thank you very much for looking into it!

andreknieriem:
@o-jcardenass what is the difference? Can you make a PR for it :)

o-jcardenass:
@andreknieriem It has some changes on how the failed decoding and frames are handled, I wanted to make a PR expecting this other issue #742 to be closed to, :/ but sadly no fix for that.

Later I will make the PR with that changes that fixed this

genge1982:
> Please try this, and tell me if it works: https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.0_debug.apk

Hi this version worked good also with my android px5 autoradio with h265 settings. With other version i must switch to h264 to work

https://github.com/andreknieriem/open-headunit/issues/696#issue-4968440183

zd3sf:
Hi guys sorry for opening this again, while H264 HW decoding worked, it quite slow at 3FPS. It used to get 60 FPS on the same hardware. 

[...]

PR #792, opened 2026-08-07
Fix/video purge 738 decode regressions

o-jcardenass:
## Why

Eight issues since 3.2.0 report the same failure: no picture with working audio, blinking, freezing, "3 fps", general lag (#749, #755, #757, #744, #767, #771, #787, #788). Reporters keep naming 3.2.0-beta2 or beta3 as the last version whose picture was usable.

`git log v.3.2.0-beta3..v.3.2.0-beta4` over `decoder/`, `AapVideo.kt`, `AapProjectionActivity.kt` and `view/` returns seven commits, and only one changes decode behaviour: `18706589`, the video half of #738. It is in beta4 and every release since, and in no earlier tag.

Two of its changes combine badly. `feedInputBuffer`'s wait for a free input buffer went from 30 attempts to 3, so 300 ms down to 30 ms, and a failure there now marks the stream corrupt instead of being ignored. At beta3 a full input queue was a bare `return` and the picture healed on the next keyframe. After #738 it cycles video focus to beg for an IDR and arms a P-frame lockout that discards everything until a keyframe is positively identified.

The trigger fires constantly: in `HUR_Log_20260805_104127_934.txt` (#757, Intel Atom) the first `Input buffer full` lands **70 ms after `Codec initialized`**, before the component has drained its first buffers. And the lockout can fail to end, because it has no timeout, no re-request, and `checkKeyframe` parses NAL headers using `settings.videoCodec` (the user's preference, default `"Auto"`) rather than the pinned codec, while `ServiceDiscoveryResponse` enforces H.265 at 1440p regardless. Measured in `HUR_Log_20260806_214208_084.txt` (#755): `fed=0, rendered=0` for **16.5 seconds** with the phone connected and still sending. In a verbose #757 log the phone delivers **~50 frame-starts per second for 34 seconds**, all acknowledged, while the reporter watches 3 fps.

Should address #755, #757, #749 and #788. Reporter confirmation is outstanding.

## What changed

`31168026` reverts six pieces of `18706589`:

- `feedInputBuffer` waits 30 attempts again, and `decode()` returns `Unit`. A frame that cannot be placed is counted in `framesDropped` and forgotten.
- `waitingForKeyframe`, `checkKeyframe()` and the deferred-request machinery are deleted. `markCorruptAndRequestRecovery()` is self-healing again: corruption is scoped to the current frame and the next flag 9 or 11 clears it. `VideoRecoveryPolicy.isDeferredRequestDue` goes with it; the throttle stays.
- `KEY_PRIORITY`, `KEY_OPERATING_RATE`, `KEY_MAX_B_FRAMES` and the later `KEY_FRAME_RATE` are removed. The MediaTek unit in #755 answers the first two with `err -1010`, `KEY_MAX_B_FRAMES` is an encoder key, and the frame-rate keys declare `settings.fpsLimit`, the user's cap rather than the negotiated rate.
- The synthetic PTS returns to the wall clock. `releaseOutputBuffer(index, true)` ignores the timestamp, so it prevented no jitter.

`checkFragmentState()` stays: orphaned-fragment detection is the part of #738 that fixed a real defect, and with the lockout gone its recovery call is scoped to one frame.

`4bafc7fd` makes the reassembly buffer grow on demand (2 MB start, same 8 MB ceiling) instead of being sized from `settings.videoCodec`, which can disagree with the negotiated codec and is not readable at construction time anyway.

`sync_stall`, the crop-rect handling, the throughput telemetry and #754's codec pinning are untouched.

## Where to focus review

- **The 30 attempts is the one judgement call.** `AapVideo.process()` runs on `AapTransport:Handler::Poll`, which also carries audio and control, so a full input queue can block those for up to 300 ms instead of 30. beta2 and beta3 shipped 30, which is why it is 30. If it costs audio or control latency, the fix is tuning toward ~10, not restoring the coupling.
- `AapVideo.ensureCapacity()` copies `position()` bytes into the new buffer and continues writing. Worth checking the ByteBuffer mode handling.
- Removing the four `MediaFormat` keys is the only part that could be a downgrade on hardware that honours them.

## Verification

CI is the build gate. Separately, tested on a UNISOC MT50_YT610E4GFPSL_U (Android 14, 8 cores, Native AA over WiFi Direct, 1080p, `c2.unisoc.*` Codec2), baseline against this branch with settings preserved across `adb install -r`:

- Build and unit tests pass on both branches.
- 3 min at 1080p/H.264: 45 throughput samples, `fed == rendered`, `dropped=0` on every one, 49 to 60 fps. Indistinguishable from baseline.
- 30 min soak with music and navigation: 368 samples, `dropped=0` throughout, zero `sync_stall`, average 58.2 fps, no drift or judder. This tests the PTS change.
- 3 min explicit H.265: 50 samples, `dropped=0`. No buffer growth needed at 1080p, so that path is unexercised.
- Ping cadence at light load: 250 samples, gaps 0.934 s to 1.111 s, median 1.0 s. Audio continuous, confirmed by listening.
- Nine decoder restarts, all clean, no smearing.

**Gap: the cost of 300 ms of input-buffer patience under load is untested.** The stress run could not be executed. That unit hard-reboots under sustained multi-core load, confirmed by `/proc/uptime` reading 62 s on recovery, five times across 8, 6, 4 and 3 spin loops and identically on both builds, so it is that unit's margin and not either branch. Testing stopped rather than power-cycling further. The ping and audio numbers above are therefore at light load only. A thermal-throttle retry is the next step and has not been run.

Also unexercised: the codec-mismatch latch needs a 2560x1440 panel (that one caps to 1080p), and the smearing probe never hit its target window across nine restarts.

## Out of scope

- The wizard, DPI and resolution regressions behind #767, #744 and #787 landed in the same window and are partly addressed by `6c477c3d`. Kept separate so reporters test one variable.
- #789 targets the same symptom by softening the focus request. Under this change the lockout it works around is gone, and the #755 log from that build shows the gentler nudge prolonging the freeze because it does not force an IDR. Worth reconciling before either lands.

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/AapProjectionActivity.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/AapProjectionActivity.kt
@@ -32,6 +32,7 @@ import com.andrerinas.openheadunit.app.SurfaceActivity
 import com.andrerinas.openheadunit.connection.CommManager
 import com.andrerinas.openheadunit.contract.KeyIntent
 import kotlinx.coroutines.launch
+import com.andrerinas.openheadunit.decoder.DecoderStopPolicy
 import com.andrerinas.openheadunit.decoder.SoftwareYuvFrameSink
 import com.andrerinas.openheadunit.decoder.VideoDecoder
 import com.andrerinas.openheadunit.decoder.VideoDimensionsListener
@@ -243,7 +244,7 @@ class AapProjectionActivity : SurfaceActivity(), IProjectionView.Callbacks, Vide
             val container = findViewById<FrameLayout>(R.id.container)
             if (::projectionView.isInitialized) {
                 videoDecoder.softwareYuvFrameSink = null
-                videoDecoder.stop("projectionViewRecreate")
+                videoDecoder.stop(DecoderStopPolicy.REASON_PROJECTION_VIEW_RECREATE)
                 container.removeView(projectionView as View)
             }
             isSurfaceSet = false
@@ -1317,7 +1318,7 @@ class AapProjectionActivity : SurfaceActivity(), IProjectionView.Callbacks, Vide
         AppLog.i("SurfaceCallback: onSurfaceDestroyed. Surface: $surface")
         isSurfaceSet = false
         commManager.send(VideoFocusEvent(gain = false, unsolicited = false))
-        videoDecoder.stop("surfaceDestroyed")
+        videoDecoder.stop(DecoderStopPolicy.REASON_SURFACE_DESTROYED)
     }
 
 
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/AapVideo.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/AapVideo.kt
@@ -8,13 +8,26 @@ import java.nio.ByteBuffer
 
 internal class AapVideo(private val videoDecoder: VideoDecoder, private val settings: Settings, private val onFrameCorrupted: () -> Unit) {
 
-    private val messageBuffer = ByteBuffer.allocate(
-        if (settings.videoCodec == VideoDecoder.CodecType.H265.settingsValue) {
-            Messages.DEF_BUFFER_LENGTH * 64 // ~8MB for H.265 support
-        } else {
-            Messages.DEF_BUFFER_LENGTH * 16 // ~2MB for H.264 legacy support
-        }
-    )
+    companion object {
+        /** Enough for H.264 at the resolutions most head units negotiate. ~2MB. */
+        private const val INITIAL_BUFFER_BYTES = Messages.DEF_BUFFER_LENGTH * 16
+
+        /** Ceiling for [ensureCapacity]. ~8MB, which covers H.265 at 4K. */
+        private const val MAX_BUFFER_BYTES = Messages.DEF_BUFFER_LENGTH * 64
+    }
+
+    /**
+     * Reassembly buffer for fragmented frames. Grows on demand.
+     *
+     * It used to be sized up front from settings.videoCodec, which is the user's preference and
+     * not the codec the session runs: ServiceDiscoveryResponse negotiates H.265 at 1440p and 4K
+     * whatever that setting says, and falls back to H.264 when the user asked for H.265 on a
+     * device with no HEVC decoder. Either mismatch handed an H.265 stream the 2MB H.264 buffer and
+     * turned every large frame into a fragment overflow. The real answer is also not available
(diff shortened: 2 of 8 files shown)


o-jcardenass:
I ran this test on my revert branch
PASS      Build + unit tests, both branches
PASS      3 min 1080p/H.264 A/B vs baseline: 45 samples, fed == rendered, dropped=0, 49-60 fps
PASS      30 min soak (music + nav): 368 samples, dropped=0, zero sync_stall, 58.2 fps avg, no drift
PASS      3 min explicit H.265: 50 samples, dropped=0 (buffer growth never needed at 1080p)
PASS      Ping cadence at light load: 250 samples, median 1.0 s, max 1.111 s; audio continuous
PASS      Nine decoder restarts: all recovered clean, no smearing
NOT RUN   Stress under load: unit hard-reboots (uptime 62 s), 5 attempts, identical on both builds
NOT RUN   Codec-mismatch latch: needs a 2560x1440 panel, this one caps to 1080p
NOT HIT   Smearing probe: target window never reached across the nine restarts

I didn't make any change to the settings wizard nor the dpi

Give it a try https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.3_videorevert.apk

andreknieriem:
Just tested in on my old shitty china tablet and it runs way better. Hoepfully the other users from the issues will have the same result, so we can fix this finally and make a new release

dd310:
> I ran this test on my revert branch
> PASS      Build + unit tests, both branches
> PASS      3 min 1080p/H.264 A/B vs baseline: 45 samples, fed == rendered, dropped=0, 49-60 fps
> PASS      30 min soak (music + nav): 368 samples, dropped=0, zero sync_stall, 58.2 fps avg, no drift
> PASS      3 min explicit H.265: 50 samples, dropped=0 (buffer growth never needed at 1080p)
> PASS      Ping cadence at light load: 250 samples, median 1.0 s, max 1.111 s; audio continuous
> PASS      Nine decoder restarts: all recovered clean, no smearing
> NOT RUN   Stress under load: unit hard-reboots (uptime 62 s), 5 attempts, identical on both builds
> NOT RUN   Codec-mismatch latch: needs a 2560x1440 panel, this one caps to 1080p
> NOT HIT   Smearing probe: target window never reached across the nine restarts
> 
> I didn't make any change to the settings wizard nor the dpi
> 
> Give it a try https://github.com/o-jcardenass/open-headunit/releases/download/v.3.2.0-fixes/OHU3.2.3_videorevert.apk
> 

Used this build all day and have experienced no issues at all.
The occasional severe lag (~3fps) that sometimes occurred right after connection seems to be gone.

Issue #744, opened 2026-07-31
Version 3.2.0 is worse than the previous version.


Farruco416:
Hello, the app worked perfectly for me up to version 3.1.1, and I consider it very good. The newer updates, 3.2.0 and all, are worse than the last. The app crashes, disconnects, and the screen resolution is worse.
Regards and best of luck.

o-jcardenass:
Hi, please share the logs from the beta that you have tried so we can take a look.
Settings -> Advanced -> Change Log Level to Verbose -> Start Capturing Logs
Go back and use the app as you normally do, and then go back to
Settings -> Advanced -> Export Logs
and reply here with the log file attached

Farruco416:
Sorry, I've reverted to version 3.1.1, which I find to be the most stable so far. Version 3.2.0 has stability issues; the auto-configuration screen always appears even after manual configuration, etc. Regards.

Farruco416:
As I said, I'm sticking with version 3.1.1 and closing this issue.

andreknieriem:
That's wrong, we need to find bugs and make it better as 3.1.1 ;) Can you please try 3.2.1 bugfix release.

Farruco416:
Attached is the log file after a clean installation of the latest version 3.2.1. It's configured in native mode, which worked correctly in version 3.1.1. Now the audio cuts out for both navigation and music. The problem occurs with both Wi-Fi and USB connections. I installed this latest version on two different Android radios, one MediaTek and one Rockchip PX5, and the problem is identical on both. Native mode doesn't work on the Rockchip PX5.

[HUR_Log_20260805_114153_005.txt](https://github.com/user-attachments/files/30738964/HUR_Log_20260805_114153_005.txt)

Farruco416:
Attached is the log file generated with version com.andrerinas.headunitrevived_3.2.1_debug.apk. The audio problems (audio cuts out, goes silent) occur when connecting via Bluetooth and Wi-Fi. When connecting via USB, the audio works correctly, but if Bluetooth is activated, the audio cuts out.

[HUR_Log_20260805_164531_740.txt](https://github.com/user-attachments/files/30749726/HUR_Log_20260805_164531_740.txt)

andreknieriem:
Are you talking about 2s and stop? This is a known Android Auto Bug for AA itself. You need to turn off Audio sink or turn off bt media.

Farruco416:
[...]

PR #810, opened 2026-08-09
Native AA wake poke takes down the head unit's own hands-free link (#744)

o-jcardenass:
Reported in #744: phone calls come out of the phone instead of the car, and the head unit shows Bluetooth disconnected while the phone shows it connected.

The Native AA wake poke connects to one of the phone's audio-gateway records to raise an ACL connection so the phone starts its wireless setup. A poke that **connects** takes the phone's single hands-free slot, and the head unit's own `HfpClientConnectionService` is dropped to make room — it logged its disconnect from the same peer 4 ms after our `socket.connect()` returned, and the link stayed down for eight minutes until a Bluetooth adapter cycle. Since v.3.2.0 the poke retries every ~30 s (`38a82373`) where v.3.1.1 tried once, so the slot is held rather than borrowed.

Reordering the targets back to v.3.1.1's is not the fix: a **successful** HSP-AG poke was measured leaving the link down for three minutes too. A phone serves HFP-AG and HSP-AG from one headset connection, so which record is asked for was never the lever.

### Fix

`BluetoothWakePolicy` (new, pure, 15 unit tests) decides when a poke may run. Three commits:

1. **Name the record in the logs.** Both UUIDs move out of `NativeAaHandshakeManager`, so a capture says `HFP-AG` rather than `0000111f-…`. One had been stored for years as `A2DP_SOURCE_UUID`, which it never was, so the values are pinned in tests. No behaviour change.
2. **Skip the poke when a hands-free link is already up** — which is also when it is pointless, that link *being* the ACL connection a poke exists to create. Units where the poke is load-bearing have no such link to read and are unaffected; an unreadable adapter still pokes, rather than silently disabling a mechanism some units cannot connect without.
3. **Don't poke an unpaired phone, and don't forget one we cannot read.** `connect()` to an unpaired device solicits pairing as a side effect. Pruning the stored Auto Start MAC gets the opposite bias: `getBondState()` answers `BOND_NONE` when the Bluetooth service is merely unavailable, so an adapter that is off looks exactly like a phone the user unpaired, and deleting on that reading is permanent.

### Verification

On hardware: the guard fires with the link up and opens no socket, Android Auto still connects in ~6.6 s with the poke suppressed, the manual poke still runs when there is no link to protect, and it refuses an unpaired device with no pairing dialog.

Unverified: the prune's adapter-off path needs the retry loop to read bond state while the adapter is genuinely disabled, and that window could not be landed. The rule is unit-tested and `isEnabled` is checked before `bondState`, so what is unconfirmed is the live race, not the decision.

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/BluetoothWakePolicy.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/BluetoothWakePolicy.kt
@@ -3,11 +3,12 @@ package com.andrerinas.openheadunit.aap
 import java.util.UUID
 
 /**
- * Which of the phone's Bluetooth records the Native AA wake poke may reach for, and what to call
- * them in a log.
+ * Rules for the Native AA wake poke: which of the phone's Bluetooth records it may reach for, and
+ * when it may connect at all.
  *
  * The poke raises an ACL connection so the phone notices a wireless-capable head unit. It is
- * load-bearing — some units never connect without it — so nothing here switches it off.
+ * load-bearing — some units never connect without it — so nothing here switches it off. What the
+ * rules decide is when a poke would cost more than it gains.
  *
  * Pure and unit-tested; the sockets live in `NativeAaHandshakeManager`.
  */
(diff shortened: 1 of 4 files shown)


andreknieriem:
Hey @o-jcardenass, have merged it, now there is an build error: e: open-headunit/app/src/main/java/com/andrerinas/openheadunit/connection/NativeAaHandshakeManager.kt:578:82 Unresolved reference: handsFreeLinkState

Did I mess up anything at merge? Because I don't see the error.

o-jcardenass:
I'll have a look on it.
Out of topic, please don't open 3.3.0-alpha yet, I have found at least 5 regressions on the native path alone

andreknieriem:
Yep, me too. Just looked at it. But not really on PC the next days, because of holiday of my daughter...

o-jcardenass:
@andreknieriem I just fixed it on #814. Sorry it was my fault hahaha

Issue #803, opened 2026-08-08
[BUG] OpenHU Native AA: Steering-wheel double-skips and CommManager debounce failure with Bluetooth connected [Steering-wheel Next/Previous buttons cause double track skips only when Android Auto is connected through OpenHU]


dhilibann-arch:
### **PLEASE DO NOT CLOSE WITHOUT READING THE ATTACHED REPORT (REQUEST)**

[Android Auto Double Track Skip Investigation.docx](https://github.com/user-attachments/files/30862134/Android.Auto.Double.Track.Skip.Investigation.docx)

[Android Auto Double-Skip Investigation.pdf](https://github.com/user-attachments/files/30862317/Android.Auto.Double-Skip.Investigation.pdf)

As the title mentions I have issue of track skipping

I have given the log and details of test I have done with respect to the issue into chat GPT and it had given me a complied issue report

Since the report is lengthy and the formatting here inside the text box becomes distorted

I have downloaded the report as docx file, and I have attached the same 

Log 1 2 3, I have given to chat GPT is also attached for the developers’ reference

Additional logs that I have not shared with chat GPT are also attached for additional references

I hope the developers won’t mind reading from a docx file

Really sorry I am unable to put those details in this text box

[1.txt](https://github.com/user-attachments/files/30862286/1.txt)
[2.txt](https://github.com/user-attachments/files/30862288/2.txt)
[3.txt](https://github.com/user-attachments/files/30862287/3.txt)

[HUR_Log_20260808_210834_918.txt](https://github.com/user-attachments/files/30862316/HUR_Log_20260808_210834_918.txt)
[HUR_Log_20260808_212007_336.txt](https://github.com/user-attachments/files/30862315/HUR_Log_20260808_212007_336.txt)
[HUR_Log_20260808_213901_518.txt](https://github.com/user-attachments/files/30862314/HUR_Log_20260808_213901_518.txt)

@andreknieriem 
@o-jcardenass

dhilibann-arch:
Issue Summary

A single physical steering wheel key press (e.g., Next/Previous track) produces duplicate logical key events (AA=87 / AA=88) at the head unit input layer.  

In working sessions (Log 2), OpenHU’s CommManager successfully suppressed the double skip via its internal debouncing logic:  

CommManager: Debouncing logical key 87 (DOWN)- dropped duplicate trigger within 536ms  

[...]

PR #815, opened 2026-08-10
Input: let the Bluetooth side keep the media buttons

o-jcardenass:
## What

Settings -> Input -> **Send Media Buttons To Android Auto** (Always / Automatic / Never), plus a rewrite of the key de-duplication in `CommManager.sendKey` as a pure, tested policy object.

Addresses #803, and the steering-wheel double-skip reported by email against a BMW F30.

## Why

One press of Next skipping two tracks is **two consumers acting on one button**, not one key sent twice.

The reporter settled it by toggling "Media audio" for the phone inside a single live projection session, both directions: A2DP on -> two skips, A2DP off -> one. Their logs show exactly one `AA=87` press/release pair per press in *both* states, so the second skip was never ours to debounce. AOSP's `BluetoothMediaBrowserService` publishes a media session whose skip-to-next issues an AVRCP passthrough FORWARD at the source device -- the phone we are projecting -- the same head-unit-as-A2DP-sink topology `d2dff1df` measured for the audio-focus pause loop.

Play/pause never doubles for the same reporter, which corroborates it: passthrough PLAY/PAUSE are state-specific and converge, FORWARD adds up. A fault inside this app would double both.

Nothing arbitrates the two consumers for us: `setConnectionPolicy`/`setActiveDevice` are `BLUETOOTH_PRIVILEGED`, `abortBroadcast()` cannot reach the framework's media-button route (single target, not an ordered fan-out), and taking audio focus to deactivate the sink's session is precisely what `d2dff1df` removed. So the lever is which consumer we leave the key to.

## The setting

| Mode | For |
|---|---|
| **Always** (default, stored `0`) | today's behaviour -- an unset preference is unchanged |
| **Automatic** | phone paired to *this* head unit: suppresses media keys while A2DP is up |
| **Never** | phone paired to the *car's own* system (the BMW case): our adapter sees no link, so Automatic is structurally blind to it |

Non-media keys are never held back in any mode. That is the BMW reporter's actual ask -- keep the iDrive controller, drop the media buttons.

Defaulting to Automatic would silently disable media keys for everyone whose phone is merely paired for calls on a unit whose Bluetooth side never touches these buttons, hence Always as the zero value.

## Also fixed, independent of #803

Three defects in `sendKey`, all reachable on `main` today:

1. `keyStates` was latched **before** the debounce returned, so a dropped press still marked the key down and its release was transmitted -- `DOWN, UP, UP` on the wire, present in the reporter's own logs.
2. A press whose release never arrived latched the key: every later press matched the held state and returned in silence until the next disconnect. This is the shape of the long-standing "steering keys stop working until I restart" reports.
3. Deliveries now dedupe on `KeyEvent.getDownTime()` where the caller has a real event, instead of guessing from elapsed time. Two deliberate presses inside 600 ms used to merge into one; they no longer do. The 600/300 ms windows survive only for the proprietary OEM broadcasts that carry no `KeyEvent`.

`sendKey` also takes a source label and every call site names itself, so a log shows how many *distinct physical presses* there were rather than how many deliveries.

### DIFF
--- a/app/src/main/java/com/andrerinas/openheadunit/AapBroadcastReceiver.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/AapBroadcastReceiver.kt
@@ -56,7 +56,9 @@ class AapBroadcastReceiver : BroadcastReceiver() {
                 intent.getParcelableExtra(KeyIntent.extraEvent)
             }
             event?.let {
-                component.commManager.sendKey(it.keyCode, it.action == KeyEvent.ACTION_DOWN)
+                component.commManager.sendKey(
+                    it.keyCode, it.action == KeyEvent.ACTION_DOWN, it.downTime, "media-key-intent"
+                )
             }
         } else if (intent.action == ProjectionActivityRequest.action){
             if (component.commManager.connectionState.value is CommManager.ConnectionState.TransportStarted) {
--- a/app/src/main/java/com/andrerinas/openheadunit/aap/AapProjectionActivity.kt
+++ b/app/src/main/java/com/andrerinas/openheadunit/aap/AapProjectionActivity.kt
@@ -1451,7 +1451,7 @@ class AapProjectionActivity : SurfaceActivity(), IProjectionView.Callbacks, Vide
         // route it through CommManager so it can be remapped and sent to Android Auto.
         if (commManager.connectionState.value is CommManager.ConnectionState.TransportStarted &&
             ProjectionKeyPolicy.shouldRouteBackKeyToProjection(cachedKeyCodes, event.keyCode)) {
-            commManager.sendKey(event.keyCode, action == KeyEvent.ACTION_DOWN)
+            commManager.sendKey(event.keyCode, action == KeyEvent.ACTION_DOWN, event.downTime, "projection")
             return true
         }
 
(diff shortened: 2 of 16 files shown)


andreknieriem:
This is still in testing right?

o-jcardenass:
Yes, I'm waiting for #803 testing

o-jcardenass:
Tested on real hardware (UNISOC head unit + phone, Native AA over WiFi Direct), keys injected via the Microntek broadcast pair (com.microntek.irkeyDown/irkeyUp), which is the only adb-reachable path that reproduces the real double delivery (src=carkey direct call + src=key-broadcast via the projection activity's receiver).

The actual point of this branch, confirmed clean:
- Automatic mode holds all three media keys back with a live A2DP link up, and forwards them cleanly once the link drops. The probe never read null on this hardware, so Automatic is a live mode here, not a fallback.
- Non-media keys (D-pad, Enter, Back) are never held back in any of the three modes, 18/18 across Always/Automatic/Never.
- The setting itself: absent key defaults to Always, persists across a force-stop, and an out-of-range value falls back to Always with no crash.

The de-dup rewrite, also confirmed:
- One physical press still produces exactly one forwarded key even with both delivery paths firing.
- A stuck key (press with no release) correctly self-cand lets the next press through.
- Two deliberate presses 100ms apart (previously merged) now both go through, since dedup is keyed on downTime instead of the old window.

PR #845, opened 2026-08-17
Nearby: build the tunnel whichever callback arrives last, and say why when it doesn't

tamer1an:
A socket read that returned empty ended the session outright: one 15 s
window with no traffic logged "connection lost" and returned -1, which
tears the transport down and costs a full handshake and the "Android is
starting" screen to recover.

That window is not evidence of a dead link. A stall that long is ordinary
on WiFi -- a scan, a power-save wake, a retransmission burst, or a head
unit busy enough not to service the socket for a moment -- while a link
that has actually gone is rare. The file's own comment above the read said
as much ("the phone can be idle for minutes and that's normal") while the
code below it assumed the opposite.

The USB path already gets this right: on a timeout it returns 0 and keeps
reading, and only gives up once the connection itself reports it is down.
That test has no equivalent here, because a TCP socket whose peer has
vanished goes on reporting isConnected == true -- the flag records that
connect() once succeeded and nothing more. So the socket path could not
tell idle from dead, and resolved it by assuming dead.

Stop inferring and ask. The protocol already carries a ping, which this
app answers for the phone but never sends. A silent window now costs a
PING_REQUEST rather than the session, and only a peer that ignores three
consecutive probes is treated as gone -- roughly a minute, against the one
window it used to take. Any inbound byte resets the count, including the
response to a probe still in flight, so a busy link never approaches the
limit and a dead one cannot escape it.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


o-jcardenass:
Hi @tamer1an, I was also doing changes on AapRead in #861 and also took a look at your PR, I think we did similar things in some parts and no so much in others, give #861 a try and see if it contains part of the solution you were looking for

andreknieriem:
First: @tamer1an Thanks for your PR! If you can check what @o-jcardenass said, I can approve the merge or let it go ;)

tamer1an:
> Hi @tamer1an, I was also doing changes on AapRead in #861 and also took a look at your PR, I think we did similar things in some parts and no so much in others, give #861 a try and see if it contains part of the solution you were looking for

@o-jcardenass @tamer1an — Thanks for catching the PR #861 overlap!

We've designed our `IdleLinkProbePolicy` to be **data-driven and testable**:
- Uses the protocol's own ping (already sent/answered elsewhere)
- Only after 3 unanswered probes do we declare the link dead (~60 s of silence)
- Any inbound traffic resets the counter

**Compared to guessing** (the old way): 60 seconds is wide enough that normal WiFi stalls never trigger a false disconnect, yet short enough that truly dead links are caught before the user notices.

**For merging**: We recommend:
1. ✅ Approve these 4 commits (each is independently valuable)
2. 📊 Test both this solution and #861 on the same hardware for 30 min
3. 🏆 Keep whichever proves more reliable

o-jcardenass:
Thanks @tamer1an I'll take a look on it :)

o-jcardenass:
@tamer1an I went through the commits and tested the branch on my head unit. It builds and the tests pass.

Going through your commits, this is where I landed:

| Commit | |
|---|---|
| `36cbad0b` idle probe | Already withdrawn by your own merge, nothing to do. |
| `40294af4` AppLog stack trace | Worth keeping, I would take it as is. Confirmed it on the device. |
| `a7fb43b4` Nearby | Worth keeping, after a few small tidies that are in the attached notes. |
| `ef3fba23` dev app | The one I would leave out, two reasons below. |
| `66e16e0b` merge | Nothing to do. |

On the dev app. Having a dev and a normal version installed side by side will confuse people, and in most languages both icons end up with exactly the same name. They also do not coexist: I ran both on one unit and whichever starts second cannot connect, while the one already running loses its session. That is not something you introduced, the app was simply never built expecting two copies of itself on one device. The same commit also changes the navigation permission, which quietly changes behaviour in release builds and would stop third party navigation apps receiving updates.

The build script should stay local. The repo already ignores a similar one, there is just no rule for `.ps1` files yet.

Both of those, the second app and the script, are really @andreknieriem's call rather than mine.

One last note, and not a criticism: the README says Android Auto 17.4+ breaks Wireless Helper, and Nearby runs under Wireless Helper, so that path may not have long left. The fix is still correct and worth merging, I would just not spend much more time hardening it.

The detail and the file references are all in the attached notes.
[pr845-findings.md](https://github.com/user-attachments/files/31314607/pr845-findings.md)

tamer1an:
> @tamer1an I went through the commits and tested the branch on my head unit. It builds and the tests pass.
> 
> Going through your commits, this is where I landed:
> 
> Commit	
> `36cbad0b` idle probe	Already withdrawn by your own merge, nothing to do.
> `40294af4` AppLog stack trace	Worth keeping, I would take it as is. Confirmed it on the device.
> `a7fb43b4` Nearby	Worth keeping, after a few small tidies that are in the attached notes.
> `ef3fba23` dev app	The one I would leave out, two reasons below.
> `66e16e0b` merge	Nothing to do.
> On the dev app. Having a dev and a normal version installed side by side will confuse people, and in most languages both icons end up with exactly the same name. They also do not coexist: I ran both on one unit and whichever starts second cannot connect, while the one already running loses its session. That is not something you introduced, the app was simply never built expecting two copies of itself on one device. The same commit also changes the navigation permission, which quietly changes behaviour in release builds and would stop third party navigation apps receiving updates.
> 
> The build script should stay local. The repo already ignores a similar one, there is just no rule for `.ps1` files yet.
> 
> Both of those, the second app and the script, are really @andreknieriem's call rather than mine.
> 
> One last note, and not a criticism: the README says Android Auto 17.4+ breaks Wireless Helper, and Nearby runs under Wireless Helper, so that path may not have long left. The fix is still correct and worth merging, I would just not spend much more time hardening it.
> 
> The detail and the file references are all in the attached notes. [pr845-findings.md](https://github.com/user-attachments/files/31314607/pr845-findings.md)

Exactly the two commits we endorsed. Committing as two logical changes.

Plus a one place I deviated from the suggested fix: for F6 their literal patch (socket != null && pending == null) is also true after a healthy completed session — it would explain failures the comment says not to explain. I used activeNearbySocket?.inputStreamWrapper == null instead, which tests whether the inbound half ever attached. I've flagged the reasoning in the draft reply so they can push back.

PR #877, opened 2026-08-22
Fix/session lifecycle and video concealment

o-jcardenass:
## Why

Two stacks that touch the same four files, so they land together: session lifecycle and link diagnostics, and the video path's response to a lost access unit.

## Wireless, VPN and USB

- **Link diagnostics.** `LinkGapMonitor` runs per series (all channels, video, audio) so a dead radio is separable from a stalled media path, and a new `UplinkStallMonitor` measures the send side, which `max_unacked` flow control can stall while control traffic stays healthy.
- **Idle is not a fault.** `shouldShowReconnecting` needs the link quiet as well as the picture, and `maybeRecoverFromDisplayStall` no longer rebuilds projection and decoder mid-outage for a fault no rebuild fixes.
- **Offline VPN.** `DummyVpnPolicy` owns the lifecycle so a mode change cannot stop a VPN it never started, teardown closes the tun descriptor rather than asking an established service to stop, and the playstore flavor gets the `VpnControl` copy the seam always implied.
- **USB sessions.** A wired connect quiesces NetworkDiscovery, Nearby, the handshake manager and the P2P group, and `initWifiMode()` refuses to arm while a wired session is live.
- **Wireless settings.** The hotspot band is the user's choice instead of ACS's, and each lever is shown on the screen that reads it.

## Video

- **The last fragment of a frame.** `decrypt` rejected every payload under two bytes, but only a message with flag bit `0x01` carries a type, so a legal 1-byte tail cost the whole frame. Measured seven for seven in a reporter's logs.
- **Freeze, do not melt.** After a lost access unit the output thread releases buffers unrendered so the surface holds the last good frame, closing on the confirmed repaired keyframe or a 3.5 s cap sized under the display-freeze teardown.
- **The corruption path reaches the focus cycle.** Wire truncation passed `escalatable=false` and got only the inert nudge; it now uses the same rendered-frame gate as the dropped-frame path, and the cycle budget refunds after five quiet minutes so a multi-hour drive is not spent by the first bad patch.
- **Input buffer attrition.** A throw between `dequeueInputBuffer` and `queueInputBuffer` leaked the index permanently, and enough of them turn into `NO_INPUT_BUFFER` timeouts the stall watchdog answers with a rebuild the codec never earned.
- **Watchdog clocks split.** `lastOutputMs` stamps on every dequeued buffer, `lastFrameRenderedMs` only on frames that reach the surface, so holding a frame off screen cannot read as a codec that stopped decoding. Behaviour preserving.
- **Log provenance.** Exported logs now name the build, device and settings at WARN from both export paths, and a zero-produce SSL unwrap prints at WARN instead of DEBUG.

## Where to focus review

The VPN teardown: `stopService()` cannot stop a `VpnService` that has called `establish()`, and the previous `startService()`-with-stop-action throws on O+ while the app is losing foreground.

The concealment window's 3.5 s cap against `ProjectionWatchdogPolicy.DISPLAY_FREEZE_MS`, which is why both constants now live in the policy with tests pinning the relation.

## Verification

Each half was built clean and unit tested on the rig, and each was validated on hardware (UNISOC MT50, Android 14, Native AA wireless, POCO X3 NFC). 738 unit tests on the branch.

Video rounds: hold/restore 2651-2683 ms across truncation, holed-run and orphan faults, one millisecond off the pre-concealment repair; a 30-fault storm produced exactly one freeze; 121 clean-control windows all `concealed=0` with zero watchdog reactions.

Wireless rounds: three minutes of a stationary Maps screen printed zero `inbound video quiet` lines at 39-46 fps, and a clean ten-minute session printed none of the new lines.

@andreknieriem Take a look on the Play Store exclusion for the dummy VPN at commit a960b9cf61a6a2acf582352e9b60f10258821163


andreknieriem:
Looks good. Will test it out

PR #885, opened 2026-08-24
3.3.0 beta1

andreknieriem:
- Begin for theming of the App.
- Refactor WiFi-Code from AapService into their own classes for better maintenance, thanks to @MrEAlderson
- USB-Blacklist Filter to prevent the app from asking for non Android phones
- Added: Option to disable the clock, thanks to @MrEAlderson
- Added: French translation 🇫🇷 thanks to @phiDu-fr
- Fix/session lifecycle and video concealment
- Video: pace the transport thread instead of shedding reference frames
- UI: a destroyed activity stops listening, and a recycled row keeps its subtitle


o-jcardenass:
@andreknieriem I'll start my checks + in device tests

o-jcardenass:
@andreknieriem @MrEAlderson Please test again, I fixed the Native Path and the Wireless Helper. Hope I didn't break your changes

@MrEAlderson  Thanks for your hard work

andreknieriem:
@o-jcardenass, do you think we can release the first beta version with all fixes and features? I mean it is beta and we can fix for the final release ;)

o-jcardenass:
@andreknieriem hahaha sorry, I didn't marked it as approved, I tested Native + Nearby + Wifi direct. So yes good to go

o-jcardenass (approved):

PR #892, opened 2026-08-25
More refactoring + Improved self mode startup

MrEAlderson:
- Self mode will now suggest missing permissions if launch failed
- Self mode will now print more helpful info if launch failed (AA not installed, not installed as system app, and that it failed in the first place)
- Restructured self and usb classes into their own packages + extracted them from AapService
- Modulized self launching mode


andreknieriem:
2 things, but I will change that ;) 

HomeFragment.kt (Line 628). 

activity?.let { act ->
    { // <-- Unnecessary, the function will never get called

and 2 strings failed_aa_not_system, failed_aa_not_installed are not translated to all languages.

PR #993, opened 2026-09-16
Measure and size the audio sink, give video its own transport thread, gate and bound Native AA bring-up, and stop old Android promising what it cannot do

o-jcardenass:
Five commits on `main` (`5ce51c5e`), 2164 unit tests. Ten hardware rounds on a UNISOC MT50 head unit (Android 14), a Samsung SM-T230 (API 19) and an HP Slate 7 Plus (API 17) as head units, with a POCO X3 NFC and a Moto edge 30 neo as the phones, those two also standing in as head units for each other.

- **Audio** (`a392f1e4`): the sink could not be sized or trusted. Its cushion came from the latency setting rather than the granted capacity, the shared mixer track was sized by whichever Media Sink Setup arrived first, a re-sent one destroyed a live sink whose codec had not changed, a full queue refused the newest chunk instead of shedding the oldest, and the queue's depth was a chunk count that on an AAC sink was derived from encoded bytes divided by a PCM frame's width. Each is fixed at its cause, the queue is measured at the decoder and reported as a delay against the ceiling in force, and a sink that keeps underrunning deepens its own cushion and says so. `AudioSinkHealthMonitor` is the instrument all of it is read with.

- **Transport** (`5e96eebe`): video decoded on the read thread that also services audio, so a slow decoder blocked audio outright, and the video media ack went out on receipt rather than after the decode, which removed the only bound the backlog ever had and let the queue climb without recovering. Video now decodes on its own thread and acks after decoding, restoring that bound, with a 256 message ceiling that sheds the oldest as a backstop and marks the fragment run holed so a partial access unit is not decoded as a whole frame. `TransportDispatchMonitor` reports the lot on one line.

- **Native AA** (`7899e35d`): the wireless bring-up could not be gated, bounded or stopped. Two bring-ups could race and delete each other's group, wireless armed whether or not the user had chosen WiFi and over an open settings screen, nothing stood the poke loop down on a unit whose own poke raises its auto-start, and below Android 10 a group the platform renames every create was never diagnosed as one. It is now gated by a create claim and the user's own choices, stoppable from the status pill until a bring-up is asked for, and the wake measures its own damage per unit instead of asking the user. Two silent hands-free conditions now speak: a stand-in hold the phone never answers, and a wake past a link this unit holds for another phone. Neither is repairable on any Android release, so both are reported rather than fought.

- **The session, and the rest** (`ba009519`, `4c35c014`): a handshake whose projection never came up held a live socket nobody read, because the raise went through a full-screen-intent notification Android 14 demotes. A foreground app now raises it directly, a deadline retries at 8 s and ends the session at 16 s, and a user exit waits for the wireless teardown to give the network back rather than leaving a group for the next create to meet as BUSY. Alongside them: the USB reconnect is no longer abandoned inside an AOA mode switch, sharing an exported log no longer kills the app below Android 10, the user's auto-start Bluetooth device is no longer cleared behind their back, and the head unit server is found by sweeping the joined network's subnet rather than whichever interface the unit lists first.


andreknieriem:
Hey I am just testing it. For me the audio stutter is gone, but it seems to the cost of the framerate. Did not have your commit from 3 Minutes ago, but my very low end china tablet always getting:

`[7536] VideoDecoder.logFeedPacing | Feed queue full - pacing the transport thread instead of shedding frames (+39 more since the last line)`

It runs never very good in the past, but now it has 1-2 seconds touch to interact offset. Sounds runs smoothly.

andreknieriem:
And another "Bug" when I close the connection. Go to the settings and change some things and click back to the main screen AA Wireless Setup starts immediately, like on App-Start. I guess this not a wanted behaviour, if you just change some settings after AA and do not want a reconnect, or how do you see that? And maybe we should add a big X Circle in the pill to cancel the wireless launch?

o-jcardenass:
> Hey I am just testing it. For me the audio stutter is gone, but it seems to the cost of the framerate. Did not have your commit from 3 Minutes ago, but my very low end china tablet always getting:
> 
> `[7536] VideoDecoder.logFeedPacing | Feed queue full - pacing the transport thread instead of shedding frames (+39 more since the last line)`
> 
> It runs never very good in the past, but now it has 1-2 seconds touch to interact offset. Sounds runs smoothly.

@andreknieriem please share me the log from your ultra lower end china table haha, maybe there is something else left to tune.

EDIT: I experience the same with my another ultra low spec tablet, and the solution was setting resolution to 480p. One step lower that "supported" 720p

PR #1001, opened 2026-09-19
Add automatic Wi-Fi Direct for Headunit Server mode

mixalbl4-127:
## Summary
Adds an opt-in **Automatically start Wi-Fi Direct** checkbox to **Headunit Server / Auto** for use with the root developer-server mode in Wireless Helper.

Companion phone-side PR: https://github.com/andreknieriem/wireless-helper/pull/70

On Android Auto 17.4+, the phone can run the developer headunit server while this app creates the OpenHU P2P group and connects directly to the phone on TCP 5277. Root is required only on the phone. The checkbox defaults to off; Manual mode keeps its existing behavior.

## Changes
- Start P2P behind conflicting-hotspot teardown, give it priority over automatic hotspot creation, and recover visibility/group availability.
- Discover the server only after a P2P client and local IPv4 interface are available. Scan the actual P2P subnet with source-address binding rather than falling back to station Wi-Fi or a fixed gateway.
- Reuse the discovered connected socket for AA transport; keep one scan in flight and pause discovery during connection/session setup.
- Separate the local P2P device name from the group's SSID, await the OpenHU rename result with a bounded fallback, and stabilize the badge against transient or out-of-order snapshots.
- Skip saved Wi-Fi IP auto-connect for this mode. Add a scoped background action for manual connection overlays while preserving full Stop/Exit semantics and blocking late re-arms.
- Distinguish unexpected Auto/P2P transport failures from explicit exits so recovery stays armed.
- Persist/export/import the checkbox and apply configuration changes to the running launcher.
- Add English/Russian strings, JVM regression tests, setup documentation and the 3.4.1 version bump.

## Validation
- The author built the debug apps and reports stable end-to-end projection on the phone/head-unit pair. The observed sequence was a connection loading screen, OpenHU availability, connection establishment, then working Android Auto projection.
- `git diff --check` and parsing/matching of the added EN/RU resource XML passed.
- JVM tests were added for P2P network/socket selection, configuration/rearm, saved-IP and failure policies, badge ordering/hysteresis, rename completion and overlay behavior.
- A clean run of `testGithubDebugUnitTest` and `lintGithubDebug` for the final revision has not been confirmed; the implementation environment has no Android/JDK toolchain.
- The complete acceptance matrix and all OEM/radio combinations are not claimed as verified.

Suggested verification: `./gradlew testGithubDebugUnitTest assembleGithubDebug lintGithubDebug` with the project's Android SDK/JDK setup.


andreknieriem:
Hey, give me some time to check this ;)

